PR #16, opened 2024-08-20
feat: snapshot testing added

imrishabh18:
I was working on this and had made some progress.
I think we can follow this snapshot testing way, like generate for all the files


tests/dfn.test.ts
```diff
@@ -1,11 +1,15 @@
...
 
 test("dfn8_w5.3mm_p1.27mm", async (t) => {
-  const { fp, logSoup } = await getTestFixture(t)
+  const { fp, logSoup, pcbToSvg } = await getTestFixture(t)
   const soup = fp.string("dfn8_w5.3mm_p1.27mm").soup()
 
   t.is(su(soup).pcb_smtpad.list().length, 8)
+
+  const svg = pcbToSvg(soup)
+  t.snapshot(svg)
```
seveibar, line 12: Using this doesn't allow me to see the svg in the preview though, we basically have to invent our own snapshot so that ava properly creates the files (or maybe there's an ava setting to make the artifact just a regular svg)

seveibar: Looked into this a bit- I don't think ava can be configured to do it- I think we should try to do the same thing you did in [tscircuit/builder](https://github.com/tscircuit/builder)

imrishabh18: I was thinking from the point of view of running the tests in the pipeline, if something breaks then this will fail as with comparing from previous snapshot

seveibar: yea that makes sense but we could just manually implement that with `looks-same`

The main issue rn is that @abhijitxy isn't able to preview what he's working on because the debug server is broken

seveibar: @imrishabh18 yea exactly, let's just drop an svg file in and later on we'll use `looks-same` to do snapshot testing so that it functionally is the same as `t.snapshot` for regressions

PR #21, opened 2024-08-25
Added some more tests

anas-sarkez:


seveibar (approved):
they look great!

there's no need to copy the `logSoup` btw, that just sends stuff to the debug server and now that we have svgs it's less useful

PR #22, opened 2024-08-25
Fixed SOIC fn 

anas-sarkez:
Me and @ShiboSoftwareDev  found a bug in the SOIC implementation where pin width is subtracted from the pitch which affected the gap between the pins and the silkscreen, The bug was complicated to figure out and I think it should be fixed now.


src/fn/soic.ts
```diff
@@ -33,6 +33,8 @@ export const extendSoicDef = (newDefaults: {
       } else if (!v.pl) {
         v.pl = v.pw! * (1.0 / 0.6)
       }
+      // Consider pin width when calculating pitch
+      v.p += v.pw
```
seveibar, line 37: nice

seveibar: hmm looking into this a bit, I think that we don't want to modify the pitch given by the user. We want to modify or fix the drawing function (if it's incorrect)

image

seveibar (changes requested):
i think modifying user parameters could be confusing, we should preserve the user-specified pitch and fix an error in the drawing function if there is one

anas-sarkez:
**I think I can do that**

anas-sarkez:
Screenshot 2024-08-26 at 2 44 38 PM
We fixed a bug that doesn't exist, it turned out it was behaving like it should. We just got confused by this test which doesn't make sense.

PR #29, opened 2024-09-03
Add silkscreen for all passive-fn for example [0402,0603...] , and add plus sign prop and silkscreen

anas-sarkez:


src/helpers/+_sign.ts
```diff
@@ -0,0 +1,32 @@
+import type { PcbSilkscreenPath } from "@tscircuit/soup"
```
seveibar, line 1: Can we make this filename a bit safer for different operating systems, we could call it plus_sign

src/helpers/passive-fn.ts
```diff
@@ -163,16 +164,65 @@ export const passive = (params: PassiveDef): AnySoupElement[] => {
 
   if (pw === undefined) throw new Error("could not infer pad width")
   if (ph === undefined) throw new Error("could not infer pad width")
+  // if (params.w && params.p && params.pw) {
+
+  const PcbSilkscreenPath: PcbSilkscreenPath[] = [
```
seveibar, line 169: Use camelCase for variable names. PascalCase is conventionally used for interfaces, types or classes

tests/output/cap footprint.svg
```diff
@@ -1,7 +1,16 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="800" height="600"><style>
               .pcb-board { fill: #000; }
+<<<<<<< HEAD
```
seveibar, line 3: This file has merge conflicts, along with some others

seveibar (changes requested):
Most svgs are broken due to meege conflicts

tests/res.test.ts
```diff
@@ -1,10 +1,11 @@
 import test from "ava"
 import { getTestFixture } from "tests/fixtures/get-test-fixture"
 import { su } from "@tscircuit/soup-util"
+import { res } from "src/fn/res"
 
 test("res_imperial0402", async (t) => {
   const { fp, snapshotSoup } = await getTestFixture(t)
-  const soup = fp.string("res_imperial0402").soup()
+  const soup = fp.string("2512").soup()
```
seveibar, line 9: Why was this changed? Imperial0402

seveibar (approved):

tests/res.test.ts
anas-sarkez: oops I missed when I was testing all passive fn

seveibar:
reverting this merge because it broke the repo publishing, @anas-sarkez make sure you are completing passing so you don't break the repo

image

PR #37, opened 2024-09-15
updated circuit-to-svg and dependency

anas-sarkez:


imrishabh18:
Looks good to me @anas-sarkez, Great work!

PR #40, opened 2024-09-21
.json() returns a string of used parameters

ShiboSoftwareDev:
image


ShiboSoftwareDev:
TODO:
- return this:
image
instead of this:
image
- maybe some default values should be default inside of the component definition like this one:
image

ShiboSoftwareDev:
image
I haven't updated this type definition, should I?

ShiboSoftwareDev:
I have updated the format and it's still failing

src/fn/bga.ts
```diff
@@ -134,5 +137,5 @@ export const bga = (params: BgaDefInput): AnySoupElement[] => {
     }
   }
 
-  return [...pads]
+  return { circuitJson: [...pads], parameters: JSON.stringify(params) }
```
seveibar, line 140: Why stringify?

```suggestion
  return { circuitJson: [...pads], parameters: params }
```

seveibar (commented):
Looks good except how parameters are stringified

ShiboSoftwareDev:
@seveibar 
image

seveibar (approved):

PR #53, opened 2024-10-05
replaced width default value from inch to mm

anas-sarkez:


seveibar (changes requested):
Can we fix the underlying issue? Anyone who specifies mils is going to run into the same issue and technically the dimensions of DIP components are originally in mils

PR #56, opened 2024-10-07
Migrate from ava to bun test

anas-sarkez:


tests/cap.test.ts
```diff
@@ -1,33 +1,16 @@
...
-import { toPinPositionString } from "./fixtures"
-import { getTestFixture } from "tests/fixtures/get-test-fixture"
-test("cap footprint", async (t) => {
-  const soup = fp().cap().imperial("0402").soup()
-  const ps = toPinPositionString(soup)
-  const { snapshotSoup } = await getTestFixture(t)
+import "bun-match-svg"
+import { test, expect } from "bun:test"
+import { convertCircuitJsonToPcbSvg } from "circuit-to-svg"
+import { fp } from "dist"
```
seveibar, line 4: we should import fp from "src/footprinter", not "dist" (otherwise you'll need to build before running every test)

seveibar (changes requested):
see comment

seveibar:
looks good except for dist import

seveibar (approved):

seveibar:
noice!

PR #59, opened 2024-10-10
update footprinter.ts and slop1.test.ts for better error message

ChiragAgg5k:
Screenshot 2024-10-10 at 11 12 45 PM

### DIFF
--- a/tests/fp-string-error.test.ts
+++ b/tests/fp-string-error.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, it } from "bun:test"
+import { getTestFixture } from "./fixtures" // Adjust path based on your structure
+
+describe("FP String Error Tests", () => {
+  it("should throw an error for invalid footprint function", async () => {
+    const { fp } = await getTestFixture("slop1")
+
+    const invalidFootprint = "invalidFunction"
+
+    try {
+      fp.string(invalidFootprint).soup()
+      expect.fail("Expected an exrror to be thrown, but no error was thrown")
+    } catch (error) {
+      expect(error).toBeInstanceOf(Error)
+
+      expect(error.message).toContain("Function not found for footprinter")
+      expect(error.message).toContain(
+        "Specify a valid function like .dip, .lr, .p, etc.",
+      )
+    }
+  })
+})
--- a/tests/fp-string-error.test.ts
+++ b/tests/fp-string-error.test.ts
@@ -1,5 +1,4 @@
 import { describe, expect, it } from "bun:test"
-import { isNotNull } from "../src/helpers/is-not-null"
 import { getTestFixture } from "./fixtures" // Adjust path based on your structure
 
 describe("FP String Error Tests", () => {
@@ -7,25 +6,12 @@ describe("FP String Error Tests", () => {
     const { fp } = await getTestFixture("slop1")
 
     const invalidFootprint = "trianglularsmt4"
-    const invalidFootprintParts = invalidFootprint
-      .split("_")
-      .map((s) => {
-        const m = s.match(/([a-z]+)([\(\d\.\+\?].*)?/)
-        const [_, fn, v] = m ?? []
-        if (v?.includes("?")) return null
-        return { fn: m?.[1]!, v: m?.[2]! }
-      })
-      .filter(isNotNull)
 
     try {
       fp.string(invalidFootprint).soup()
       expect.fail("Expected an exrror to be thrown, but no error was thrown")
     } catch (error) {
       expect(error).toBeInstanceOf(Error)
-
-      expect(error.message).toEqual(
-        `Function not found for footprinter "${invalidFootprintParts[0]?.fn}". Specify a valid function like .dip, .lr, .p, etc.`,
-      )
     }
   })
 })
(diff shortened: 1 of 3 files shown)


tests/slop/slop1.test.ts
```diff
@@ -47,5 +47,22 @@ describe("Slop Tests", () => {
...
+        expect.fail("Expected an exrror to be thrown, but no error was thrown")
+      } catch (error) {
+        expect(error).toBeInstanceOf(Error)
+
+        expect(error.message).toContain("Function not found for footprinter")
+        expect(error.message).toContain(
+          "Specify a valid function like .dip, .lr, .p, etc.",
+        )
+      }
+    })
```
seveibar, line 67: why did you add this to a slop test? can you move it to a different test file? Also you should request me for review to  get a review faster

seveibar (changes requested):

ChiragAgg5k:
@seveibar can you tell me which file should i add it to? thanks

seveibar:
@ChiragAgg5k a new file, you could call it fp-string-error.test.ts

ChiragAgg5k:
@seveibar can u check now?

tests/fp-string-error.test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+    const invalidFootprint = "invalidFunction"
+
+    try {
+      fp.string(invalidFootprint).soup()
+      expect.fail("Expected an exrror to be thrown, but no error was thrown")
+    } catch (error) {
+      expect(error).toBeInstanceOf(Error)
+
+      expect(error.message).toContain("Function not found for footprinter")
```
seveibar, line 16: You have to test that “invalidFunction” appears in the error output, that is the key part of the implementation that you added, so you should test it

tests/slop/slop1.test.ts
```diff
@@ -1,6 +1,6 @@
-import { describe, it, expect } from "bun:test" // Bun's test utilities
-import { getTestFixture } from "../fixtures" // Adjust path based on your structure
+import { describe, expect, it } from "bun:test" // Bun's test utilities
 import type { AnyCircuitElement } from "circuit-json"
+import { getTestFixture } from "../fixtures" // Adjust path based on your structure
```
seveibar, line 3: Revert the changes to this file (eg with git chrckout origin/main — path/to/file.ts), also in this file and the other file remove the ai generated comments

tests/fp-string-error.test.ts
seveibar: Better to just test an exact match for the whole message this test is hard to read. Just tell ne as the reviewer the full message with a toEqual assertion

seveibar (changes requested):

ChiragAgg5k:
I couldn't find a helper function that converts the input function string to its equivalent form. For eg. "invalidFunction" to "invalid". Should I use regex in the testfile itself?

ChiragAgg5k:
Screenshot 2024-10-11 at 9 53 08 AM

@seveibar

seveibar:
@ChiragAgg5k i dont understand. The issue is the user types something like fp.string(“trianglularsmt4”) and we want to tell them that either “triangularsmt4” is not a function or (also acceptable) “triangularsmt” is not a function. Whatever error message it produces should be deterministic

ChiragAgg5k:
well i am confused if user wrong "triangularSmt4", then the target.fn returns "triangular", which would not match with user's original input

ChiragAgg5k:
@seveibar do check now

seveibar (changes requested):
I have no idea what is going on with your test. Just run fp.string("traingularsmt4") and expect an error. No loops or other complexity. Why do you need a test fixture.

Also fix typos

tests/fp-string-error.test.ts
```diff
@@ -0,0 +1,31 @@
...
+      .filter(isNotNull)
+
+    try {
+      fp.string(invalidFootprint).soup()
+      expect.fail("Expected an exrror to be thrown, but no error was thrown")
+    } catch (error) {
+      expect(error).toBeInstanceOf(Error)
+
+      expect(error.message).toEqual(
+        `Function not found for footprinter "${invalidFootprintParts[0]?.fn}". Specify a valid function like .dip, .lr, .p, etc.`,
```
seveibar, line 27: This line. It should be a fixed error. No interpolation allowed

seveibar (changes requested):

ChiragAgg5k:
@seveibar sorry my bad. now i am just running the invalid case and expecting error. looks good?

tests/fp-string-error.test.ts
```diff
@@ -0,0 +1,17 @@
+import { describe, expect, it } from "bun:test"
+import { getTestFixture } from "./fixtures" // Adjust path based on your structure
+
+describe("FP String Error Tests", () => {
+  it("should throw an error for invalid footprint function", async () => {
+    const { fp } = await getTestFixture("slop1")
```
seveibar, line 6:
```suggestion
    const { fp } = await getTestFixture("fp string error with invalid fn")
```

tests/fp-string-error.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+    const invalidFootprint = "trianglularsmt4"
+
+    try {
+      fp.string(invalidFootprint).soup()
+      expect.fail("Expected an exrror to be thrown, but no error was thrown")
```
seveibar, line 12:
```suggestion
      expect.fail("Expected an error to be thrown, but no error was thrown")
```

seveibar:
You’re still not testing for the error message. As the reviewer i need to see the error message to confirm that it is properly erroring.

PR #60, opened 2024-10-11
adding types for parameters returned using .json()

### LINKED ISSUE #41: types for parameters returned using .json() method
image

bbland1:
fixes #41

### DIFF
--- a/src/helpers/zod/AnyParameterDefinition.ts
+++ b/src/helpers/zod/AnyParameterDefinition.ts
@@ -21,7 +21,7 @@ import { ssop_def } from "src/fn/ssop"
 import { tssop_def } from "src/fn/tssop"
 import { passive_def } from "../passive-fn"
 
-export const anyParameterType = z.union([
+export const any_footprinter_def = z.union([
   axial_def,
   bga_def,
   dfn_def,
@@ -45,4 +45,4 @@ export const anyParameterType = z.union([
   passive_def,
 ])
 
-export type AnyParameterDefinition = z.infer<typeof anyParameterType>
+export type AnyParameterDefinition = z.infer<typeof any_footprinter_def>
--- a/src/helpers/zod/AnyParamterDefinition.ts
+++ b/src/helpers/zod/AnyParamterDefinition.ts
@@ -19,7 +19,7 @@ import { sot563_def } from "src/fn/sot563"
 import { sot723_def } from "src/fn/sot723"
 import { ssop_def } from "src/fn/ssop"
 import { tssop_def } from "src/fn/tssop"
-import { passive_def } from "../passive-fn";
+import { passive_def } from "../passive-fn"
 
 export const anyParameterType = z.union([
   axial_def,
@@ -42,7 +42,7 @@ export const anyParameterType = z.union([
   sot723_def,
   ssop_def,
   tssop_def,
-  passive_def
+  passive_def,
 ])
 
 export type AnyParameterDefinition = z.infer<typeof anyParameterType>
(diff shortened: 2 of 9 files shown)


src/helpers/zod/AnyParamterDefinition.ts
```diff
@@ -0,0 +1,48 @@
...
+import { sot23_def } from "src/fn/sot23"
+import { sot236_def } from "src/fn/sot236"
+import { sot363_def } from "src/fn/sot363"
+import { sot563_def } from "src/fn/sot563"
+import { sot723_def } from "src/fn/sot723"
+import { ssop_def } from "src/fn/ssop"
+import { tssop_def } from "src/fn/tssop"
+import { passive_def } from "../passive-fn"
+
+export const anyParameterType = z.union([
```
seveibar, line 24:
```suggestion
export const anyDefInput = z.union([
```

seveibar (changes requested):
the is very good but it's not quite right. You need to use `z.infer` or `z.output` to represent the output of `parse`

You could use `AnyFootprintDefOutput = z.infer<any_def>` (or `z.output`)

Nice work! Requesting changes just to fix the inference issue

bbland1:
> the is very good but it's not quite right. You need to use `z.infer` or `z.output` to represent the output of `parse`
> 
> You could use `AnyFootprintDefOutput = z.infer<any_def>` (or `z.output`)
> 
> Nice work! Requesting changes just to fix the inference issue

@seveibar is it the `typeof` within the `infer`. The name change def get but I think I am a bit loss on the inference? Looking into zod and the way `AnyCircuitElement` was established in [any_circuit_element](https://github.com/tscircuit/circuit-json/blob/main/src/any_circuit_element.ts) I was using a similar structure, but I'm not sure  if that's what you are referencing here 🤔

```typescript
export type AnyParameterDefinition = z.infer<typeof any_footprinter_def>

```

seveibar:
@bbland1 it's fine now that you've switched to `z.infer`, but it can be good to be pretty specific because people often get this stuff confused.

- `z.infer`, same as `z.output`. Represents the output after parsing
- `z.input`, represents the input of the parse method

```ts
import { z } from "zod";

const userSchema = z.object({
  id: z.string().uuid(),
  email: z.string().email(),
  birthYear: z.number().min(1900).max(new Date().getFullYear()),
  role: z.enum(["user", "admin"]).default("user"),
  lastLoginTimestamp: z.string().datetime().optional()
    .transform(str => new Date(str)),
});

type UserInput = z.input<typeof userSchema>;
type UserOutput = z.infer<typeof userSchema>;

// UserInput type:
// {
//   id: string;
//   email: string;
//   birthYear: number;
//   role?: "user" | "admin" | undefined;
//   lastLoginTimestamp?: string | undefined;
// }

// UserOutput type:
// {
//   id: string;
//   email: string;
//   birthYear: number;
//   role: "user" | "admin";
//   lastLoginTimestamp?: Date | undefined;
// }
```

src/helpers/zod/AnyParameterDefinition.ts
```diff
@@ -0,0 +1,48 @@
...
+  sot236_def,
+  sot363_def,
+  sot563_def,
+  sot723_def,
+  ssop_def,
+  tssop_def,
+  passive_def,
+])
+
+export type AnyParameterDefinition = z.infer<typeof any_footprinter_def>
```
seveibar, line 48: I think we should be more specific here, Parameter and Definition are very general terms. We can also match the name of the zod type

```suggestion
export type AnyFootprinterDefinitionOutput = z.infer<typeof any_footprinter_def>
```

seveibar (approved):
looks good, see comment with proposed rename for type

bbland1:
@seveibar got it! I definitely see the needing to be clearer with names and where I got a bit lost, also where the distinction is. 

Those changes should be made, I don't have write access here if there are no other changes to merge it.

Issue #31, opened 2024-09-04
Implement silkscreen reference designators for every component


seveibar:
(feel free to edit this issue with a checklist)

Below shows a reference designator from kicad. All of our footprints should have reference designators (silkscreen_text) with the text "{REF}"
![image](https://github.com/user-attachments/assets/c9c037ee-e47a-4832-93ee-6f938bf5345c)

PR #69, opened 2024-10-20
Implement silkscreen reference for every component and added silkscreenRef function to helpers 

anas-sarkez:


anas-sarkez:
Screenshot 2024-10-21 at 1 46 33 AM
Screenshot 2024-10-21 at 12 28 26 AM

seveibar (approved):

anas-sarkez:
#31

PR #71, opened 2024-10-21
Supported holes for stampreceiver and modified pad length and holes position of stampboard

anas-sarkez:


src/fn/stampboard.ts
```diff
@@ -165,7 +165,7 @@ export const stampboard = (
         holes.push(
           platedhole(
             i + 1,
-            params.w / 2 - params.pl + outerDiameter / 2,
+            params.w / 2 - 1.61,
```
seveibar, line 168: Is this intentionally hard coded? Shouldnt it be derived or a default?

seveibar:
We can keep it hard coded if we create a CONSTANT that explains what it is eg TOP_BOTTOM_RECESSION_OFFSET

src/fn/stampboard.ts
anas-sarkez: Screenshot 2024-10-21 at 11 47 49 PM
I write it hard coded to be like the data sheet
if I need it to be dynamic then I need to modify `params.pl + outerDiameter / 2`

seveibar: oh we just need to give a name to that distance and set the default to 1.61. That parameter should be called `innerholeedgedistance` i think

anas-sarkez: Screenshot 2024-10-22 at 12 00 15 AM

seveibar: yep!

seveibar (approved):

PR #83, opened 2024-11-19
Refactor all footprinter functions by updating its scheme, all footprints now have default value of num_pins

### LINKED ISSUE #75: Any footprint should have a default number of pins; I shouldn't have to type 'tssop8' to get a footprint just 'tssop'
- footprints with a set number of pins like sot235 should not have a num_pins param as it is not configurable, while footprints like soic should have num_pins as optional with a default number of pins.

- I want to be able to type stampboard/stampreceiver and see a default footprint without having to specify the number of pins

Image

Image

anas-sarkez:
/claim #75
/closes #75


src/fn/axial.ts
```diff
@@ -11,7 +11,7 @@ import { silkscreenRef, type SilkscreenRef } from "../helpers/silkscreenRef"
 
 export const axial_def = z.object({
   fn: z.string(),
-  num_pins: z.any().transform(() => 2),
+  num_pins: z.literal(2).default(2),
```
seveibar, line 14: Hmm i think this needs optional as well?

src/fn/bga.ts
```diff
@@ -1,7 +1,7 @@
 import type { AnySoupElement, PCBSMTPad } from "@tscircuit/soup"
 import { rectpad } from "../helpers/rectpad"
 import { ALPHABET } from "../helpers/zod/ALPHABET"
-import { z } from "zod"
+import { number, z } from "zod"
```
seveibar, line 4: Just use z.number dont import number

seveibar (commented):
2 main things,

when using default i think you need optional iirc

lets not import number, it overlaps the typescript type

src/fn/axial.ts
anas-sarkez: I think we need it to be read only so we don't have to add optional. and literal without default gives me errors

seveibar: Ah ok sounds good

src/fn/bga.ts
seveibar: (Bump i think)

anas-sarkez:
> 2 main things,
> 
> when using default i think you need optional iirc
> 
> lets not import number, it overlaps the typescript type

check last commit

src/fn/pinrow.ts
```diff
@@ -1,11 +1,11 @@
-import { z } from "zod"
+import { number, z } from "zod"
```
seveibar, line 1:
```suggestion
import { z } from "zod"
```

seveibar (approved):

PR #98, opened 2024-12-18
Introduced new footprint type: `breakoutheaders`

### LINKED ISSUE #97: New footprint type: "breakoutheaders"
These are two columns of headers

AnasSarkiz:
/claim #97
/closes #97


Abse2001 (approved):
I think this looks good for me.

src/fn/breakoutheaders.ts
```diff
@@ -0,0 +1,238 @@
...
+  if (params.right) {
+    return params.right * params.p
+  }
+
+  return 51
+}
+type Point = { x: number; y: number }
+type Direction = "left" | "right" | "top" | "bottom"
+
+const getTriangleDir = (
```
seveibar, line 48: this doesn't return a direction, but it sounds like it returns a triangle direction.

`getTrianglePath` with named parameters would be better!

seveibar (commented):
minor naming issues, looks great tho!

src/fn/breakoutheaders.ts
```diff
@@ -0,0 +1,239 @@
...
+  const height = params.h ?? getHeight(params)
+  const holes: PcbPlatedHole[] = []
+  const innerDiameter = params.id
+  const outerDiameter = params.od
+  let routes: { x: number; y: number }[] = []
+  if (params.right) {
+    const yoff = -((params.right - 1) / 2) * params.p
+    for (let i = 0; i < params.right; i++) {
+      if (i === 0 && !params.left && !params.bottom) {
+        routes = getTrianglePath(
```
seveibar, line 101:
```suggestion
        silkscreenTriangleRoutes = getTrianglePath(
```

seveibar (approved):

Issue #106, opened 2025-01-05
legsoutside doesn't work for components except soic


ShiboSoftwareDev:
setting it true or false has no effect

ShiboSoftwareDev:
You can test this by calling the .params() method

Anshgrover23:
@seveibar sorry i was asking something else? the message sended wrong the thing that i want to ask is not sent y mistake?

PR #108, opened 2025-01-07
debuged the footprints legoutside, added quad legoutside default `false`

Abse2001:
/claim #106 

quad legoutside default `false` will fix the quad in snippet not having legsoutside checkbox.

CC @seveibar @ShiboSoftwareDev 
Not all footprints have both "legsOutside" and "no legsOutside" variants. Some, like QFP and QFN, are variants but differ in footprint designs (QFP with legsOutside and QFN with no legsoutside ). Others, such as MS012 and MS013 (extensions of SOIC), exist only with a legsOutside configuration.
so we will need to find a way or add parameter that say if this has a variant or not?
we can merge this and create an issue with what we should do.
image
image

### DIFF
--- a/src/fn/ms012.ts
+++ b/src/fn/ms012.ts
@@ -1,4 +1,4 @@
-import type { AnySoupElement } from "circuit-json"
+import type { AnyCircuitElement } from "circuit-json"
 import { extendSoicDef, soicWithoutParsing } from "./soic"
 import type { z } from "zod"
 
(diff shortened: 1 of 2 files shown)


ShiboSoftwareDev:
Looks like it's just another read-only param, in the future we should find a way to handle these, currently they just appear as options but have no effect because I couldn't find a way to identify a read-only param

seveibar (approved):

PR #110, opened 2025-01-07
add sot457 footprint

### LINKED ISSUE #101: transistor footprint

techmannih:
fixes #101 
/claim #101

### DIFF
--- a/src/fn/sot457.ts
+++ b/src/fn/sot457.ts
@@ -18,25 +18,26 @@ export const sot457_def = z.object({
   p: z.string().default("0.95mm"),
 })
 
-export const sot457_wave = z.object({
-  fn: z.literal("sot457"),
-  num_pins: z.literal(6).default(6),
-  h: z.string().default("3mm"),
-  w: z.string().default("4mm"),
-  pillr: z.string().default("0.225mm"),
-  pillh: z.string().default("0.45mm"),
-  pillw: z.string().default("1.45mm"),
-  pl: z.string().default("1.45mm"),
-  pw: z.string().default("1.5mm"),
-  p: z.string().default("1.475mm"),
-  wave: z.boolean().optional(),
-  reflow: z.boolean().optional(),
-})  .transform((a) => ({
-  ...a,
-  wave: a.wave ?? (a.reflow === undefined ? true : !a.reflow),
-  reflow: a.reflow ?? (a.wave === undefined ? true : !a.wave),
-}));
-
+export const sot457_wave = z
+  .object({
+    fn: z.literal("sot457"),
+    num_pins: z.literal(6).default(6),
+    h: z.string().default("3mm"),
+    w: z.string().default("4mm"),
+    pillr: z.string().default("0.225mm"),
+    pillh: z.string().default("0.45mm"),
+    pillw: z.string().default("1.45mm"),
+    pl: z.string().default("1.45mm"),
+    pw: z.string().default("1.5mm"),
+    p: z.string().default("1.475mm"),
+    wave: z.boolean().optional(),
+    reflow: z.boolean().optional(),
+  })
+  .transform((a) => ({
+    ...a,
+    wave: a.wave ?? (a.reflow === undefined ? true : !a.reflow),
+    reflow: a.reflow ?? (a.wave === undefined ? true : !a.wave),
+  }))
 
 export const sot457 = (
   raw_params: z.input<typeof sot457_def>,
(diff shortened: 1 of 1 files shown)


Abse2001 (changes requested):
looking good jsut one thing you need to add an arrow that point to the first pin
image
you can check closed PRs to see how they did it.

Abse2001:
@techmannih where did you find the lp in the data sheet I can see Lp which is 0.6 not 0.67?
image

seveibar (approved):

seveibar:
hmm hold on, i think you might have used leg dimensions instead of recommended footprint dimensions

seveibar (changes requested):
make sure to use the recommended footprint, not the chip pins, it won't be manufactured properly

seveibar:
make sure to have _reflow (default) and _wave variants

package.json
```diff
@@ -36,6 +36,6 @@
     "zod": "^3.23.8"
   },
   "peerDependencies": {
-    "circuit-json": "*"
+    "circuit-json": "^0.0.130"
```
seveibar, line 39:
```suggestion
    "circuit-json": "*"
```

seveibar: you only update the version in the devDependency, not the peerDependency

src/helpers/pillpad.ts
```diff
@@ -0,0 +1,25 @@
+import type { PCBSMTPad } from "circuit-json"
+
+export const pillpad = (
+  pn: number | Array<string | number>,
+  x: number,
+  y: number,
+  w: number,
+  h: number,
+  radius: number,
```
seveibar, line 9: too many parameters, make it an object (we have a 2 parameter rule, this codebase is bad but it's only going to get worse if we don't enforce the two parameter rule)

https://github.com/tscircuit/handbook/blob/main/guides/code.md#1-two-parameter-rule

tests/sot457.test.ts
```diff
@@ -0,0 +1,22 @@
...
+test("sot457 - reflow (default)", () => {
+  const soup = fp.string("sot457").circuitJson()
+  const svgContent = convertCircuitJsonToPcbSvg(soup)
+  expect(svgContent).toMatchSvgSnapshot(import.meta.path, "sot457")
+})
+
+test("sot457 - wave", () => {
+  const soup = fp
+    .string(
+      "sot457_wave_p1.475mm_pillh0.45mm_pillw1.45mm_pillr0.225mm_pw1.5mm_pl1.45mm_h3.0mm",
```
seveibar, line 14: just `sot457_wave` should work

seveibar (changes requested):
sot457 is a standard, the user is not supposed to have to specify a million parameters. If they say `sot457` we default to `sot457_reflow`.

Make a test for `sot457_reflow` and `sot457_wave` and `sot457`. No other parameters should be supplied. The pad width and other parameters SHOULD NOT BE CUSTOMIZABLE

seveibar:
@techmannih btw good work on this, appreciate you digging in and getting it done- it's really great how much you've built to make this happen!!!!!

src/fn/sot457.ts
```diff
@@ -0,0 +1,198 @@
...
+  h: z.string().default("3mm"),
+  w: z.string().default("4mm"),
+  wave: z.boolean().default(true),
+  pillr: z.string().default("0.225mm"),
+  pillh: z.string().default("0.45mm"),
+  pillw: z.string().default("1.45mm"),
+  pl: z.string().default("1.45mm"),
+  pw: z.string().default("1.5mm"),
+  p: z.string().default("1.475mm"),
+})
```
seveibar, line 32:
```suggestion
export const sot457_wave = z.object({
  fn: z.literal("sot457").default("sot457"),
  wave: z.boolean().optional(),
  reflow: z.boolean().optional()
}).transform(a => ({
  ...a,
  wave: a.wave ?? !a.reflow ?? false,
  reflow: a.reflow ?? !a.wave ?? true,
})
```

src/fn/sot457.ts
```diff
@@ -0,0 +1,198 @@
...
+}
+
+export const sot457WithoutParsing = (
+  parameters: z.infer<typeof sot457_def>,
+): AnyCircuitElement[] => {
+  const pads: AnyCircuitElement[] = []
+
+  if (parameters.wave) {
+    for (let i = 1; i <= parameters.num_pins; i++) {
+      if (i === 3 || i === 4) {
```
seveibar, line 77: this is basically unreadable, find a different way to do this, perhaps with an object e.g.

```
const pinPositions = {
  1: ({ pw, ph }) => pillpad(...)
}
```

seveibar (changes requested):
see comments

src/fn/sot457.ts
```diff
@@ -0,0 +1,231 @@
...
+    reflow: a.reflow ?? (a.wave === undefined ? true : !a.wave),
+  }))
+
+export const sot457 = (
+  raw_params: z.input<typeof sot457_def>,
+): { circuitJson: AnyCircuitElement[]; parameters: any } => {
+  if (raw_params?.wave) {
+    raw_params.fn = "sot457"
+    const parameters = sot457_wave.parse(raw_params)
+    console.log("SOT-457 Wave Parameters:", parameters)
```
seveibar, line 48:
```suggestion
```

src/fn/sot457.ts
```diff
@@ -0,0 +1,231 @@
...
+    fn: z.literal("sot457"),
+    num_pins: z.literal(6).default(6),
+    h: z.string().default("3mm"),
+    w: z.string().default("4mm"),
+    pillr: z.string().default("0.225mm"),
+    pillh: z.string().default("0.45mm"),
+    pillw: z.string().default("1.45mm"),
+    pl: z.string().default("1.45mm"),
+    pw: z.string().default("1.5mm"),
+    p: z.string().default("1.475mm"),
```
seveibar, line 32: just remove all of these parameters. they're not parameters because the user isn't allowed to change them, use constants instead

src/fn/sot457.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  if (pn === 1) return { x: -w / 2 - 0.1, y: p }
+  if (pn === 2) return { x: -w / 2 - 0.1, y: 0 }
+  if (pn === 3) return { x: -w / 2 - 0.1, y: -p }
+  if (pn === 4) return { x: w / 2 + 0.1, y: -p }
+  if (pn === 5) return { x: w / 2 + 0.1, y: 0 }
+  if (pn === 6) return { x: w / 2 + 0.1, y: p }
+
+  throw new Error("Invalid pin number")
+}
```
seveibar, line 76: this is much better

src/fn/sot457.ts
```diff
@@ -0,0 +1,231 @@
...
+        rectpad(
+          i,
+          x,
+          y,
+          Number.parseFloat(parameters.pl),
+          Number.parseFloat(parameters.pw),
+        ),
+      )
+    }
+  }
```
seveibar, line 162: im sorry this is still a really bad function. Make this way way more readable. Use DRY principles.

seveibar: there's so much repeated parsing etc.

tests/sot457.test.ts
```diff
@@ -0,0 +1,14 @@
+import { test, expect } from "bun:test"
+import { convertCircuitJsonToPcbSvg } from "circuit-to-svg"
+import { fp } from "../src/footprinter"
+test("sot457 - reflow (default)", () => {
+  const soup = fp.string("sot457").circuitJson()
+  const svgContent = convertCircuitJsonToPcbSvg(soup)
+  expect(svgContent).toMatchSvgSnapshot(import.meta.path, "sot457")
+})
```
seveibar, line 8: why don't you test reflow as well?

seveibar: like, just test each case.

seveibar: there are 3 cases

seveibar (changes requested):
really hard to read code, lots of unnecessary repetition, variables that just shouldn't be defined as parameters

PR #111, opened 2025-01-17
getFootprintNamesByType returns {passiveFootprints, normalFootprints}

ShiboSoftwareDev:


src/footprinter.ts
```diff
@@ -119,8 +119,24 @@ export const string = (def: string): Footprinter => {
   return fp
 }
 
-export const getFootprintNames = (): string[] => {
-  return Object.keys(FOOTPRINT_FN)
+export const getFootprintNames = (): {
```
seveibar, line 122:
```suggestion
export const getFootprintNamesByType = (): {
```

seveibar (approved):
discussed at 1:56:00 on stream https://x.com/seveibar/status/1880310417056227610

PR #115, opened 2025-01-22
Feat: Introduce pad for a simple single-pin pad

### LINKED ISSUE #114: Introduce `pad` for a simple single-pin pad
Should have width and height

Make sure to test with snapshots

/bounty $10

Ayushjhawar8:
- add single-pin pad footprint
- Add pad footprint with configurable width and height
- Add snapshot tests verifying dimensions
- Position reference text above pad
- Integrate with footprinter system

/claim #114 
Closes #114


Anshgrover23 (changes requested):
@Ayushjhawar8 generate the bun snapshot see the other tests files.

Ayushjhawar8:
image

Ayushjhawar8:
@seveibar Hello, Please check 🙂

.github/workflows/bun-formatcheck.yml
```diff
@@ -23,4 +22,5 @@ jobs:
         run: bun install
 
       - name: Run format check
-        run: bunx biome format .
+        run: bun run format
```
Anshgrover23, line 25: @Ayushjhawar8 why u changed workflow.

Anshgrover23 (changes requested):

tests/__snapshots__/pad.test.ts.snap
Anshgrover23, line 1: why this .snap file i think remove this file.

Anshgrover23 (changes requested):

tests/pad.test.ts
```diff
@@ -0,0 +1,17 @@
+import { expect, test } from "bun:test"
+import { fp } from "../src/footprinter"
+import { getTestFixture } from "./fixtures/get-test-fixture"
+
+test("pad footprint", async () => {
+  const { snapshotSoup } = await getTestFixture("pad")
+  const soup = fp().pad().w(2).h(1).circuitJson()
+  expect(soup).toMatchSnapshot()
```
seveibar, line 8: don't use `toMatchSnapshot`, you can use `toMatchInlineSnapshot` but in this case `snapshotSoup` is better

.github/workflows/bun-formatcheck.yml
Ayushjhawar8: I changed it to `bun run format` because I thought it was more appropriate for the formatting setup, but if it's not related to the issue, I can revert it.

tests/__snapshots__/pad.test.ts.snap
Ayushjhawar8: I think the .snap files are automatically generated test fixtures that helped me verify if the pad footprint is generating the correct PCB properties and dimensions. If we remove them, it could break the test suite and make it harder to catch bugs when making changes to the pad footprint generator.

Ayushjhawar8: But if we still want, I can remove them completely.

Ayushjhawar8:
tested it again 
image

Ayushjhawar8:
tested it again after changes
image

.github/workflows/bun-formatcheck.yml
```diff
@@ -23,4 +23,4 @@ jobs:
         run: bun install
 
       - name: Run format check
-        run: bunx biome format .
+        run: bunx biome format .
```
Anshgrover23, line 26: @Ayushjhawar8 still why are u changing this workflow file. remove this from commit.

Anshgrover23 (changes requested):

package.json
```diff
@@ -24,6 +24,7 @@
     "@types/bun": "^1.1.10",
     "@types/node": "^20.12.13",
     "bun-match-svg": "^0.0.6",
+    "bun-types": "^1.2.0",
```
seveibar, line 27: shouldn't you just update the existing `@types/bun` dependency?

tsconfig.json
```diff
@@ -19,7 +19,8 @@
     // We are not transpiling, so preserve our source code and do not emit files
     "module": "esnext",
     "noEmit": true,
-    "lib": ["esnext"]
+    "lib": ["esnext"],
+    "types": ["bun-types"]
```
seveibar, line 23: i'm pretty sure you don't need this

seveibar (approved):
can you fix the @types/bun + bun-types issue? We shouldn't need both

otherwise looks good, you can ping @imrishabh18 to help merge b/c i might be sleeping!

Ayushjhawar8:
@imrishabh18 Please check and merge if possible 😃

imrishabh18:
Thanks for your contribution @Ayushjhawar8!

Issue #119, opened 2025-01-28
implement to92 a common transistor footprint


ShiboSoftwareDev:

ShiboSoftwareDev:
ai likes to use it

kom-senapati:
How to implement new footprint 🤔 and where to see for footprints in internet 👀

PR #123, opened 2025-02-04
Added Footprint To-92

Rishikesh63:
/claim #119


seveibar (changes requested):
Silkscreen isnt centered 
image

Rishikesh63:
> @seveibar If I am now wrong, to92 has two types. what do you think? image image

yes one is Low Profile Miniature TO-92 and other is Low Profile TO-92

Rishikesh63:
@seveibar  this Silkscreen is fine or need any modification
image

seveibar:
Can you introduce both `to92_inline` and make `to92` have the holes spaced in a triangle like it is in the [kicad viewer](https://tscircuit.github.io/kicad-viewer/#Package_TO_SOT_THT.pretty/TO-92.kicad_mod)

Also the hole widths are not correct, if we merged this our assemblers would not be able to create the board if we ordered one with it


https://github.com/user-attachments/assets/37f3b666-242b-49c9-8bbf-1efd10794209

Rishikesh63:
> Can you introduce both `to92_inline` and make `to92` have the holes spaced in a triangle like it is in the [kicad viewer](https://tscircuit.github.io/kicad-viewer/#Package_TO_SOT_THT.pretty/TO-92.kicad_mod)
> 
> Also the hole widths are not correct, if we merged this our assemblers would not be able to create the board if we ordered one with it
> 
>  Screenshare.-.2025-02-04.11_43_05.AM.mp4

thanks but i have a doubt horizontal TO-92 will be fine or i need to make like this 
image

Rishikesh63:
@seveibar  I made both to92 and to92_inline, is there need any modification

seveibar (approved):
Looks good!

PR #125, opened 2025-02-04
Fixed Bga origin parameter

Rishikesh63:
/claim #107


seveibar (commented):
it looks like because of the movement you lost the {REF}?

seveibar:
to know that this works, i'd need to see a small inline snapshot via `.toMatchInlineSnapshot` with bga coordinates from the output circuit json on one of the small bgas, you probably need to add a test that shows this working and includes `.toMatchInlineSnapshot`

Rishikesh63:
@seveibar  is it good or need any modification

tests/bga.test.ts
```diff
@@ -35,10 +35,76 @@ test("bga64_w10_h10_grid8x8_p1.27mm", () => {
...
+  // Extract pin coordinates
+  const pinData = soup
+    .filter((item) => item.type === "pcb_smtpad")
+    .map(({ x, y, number }) => ({
+      x: x.toFixed(2) + "mm",
+      y: y.toFixed(2) + "mm",
+      number,
+    }))
+
+  console.log("BGA 3x3 Coordinates:", JSON.stringify(pinData, null, 2))
```
seveibar, line 58:
```suggestion
```

tests/bga.test.ts
```diff
@@ -35,10 +35,76 @@ test("bga64_w10_h10_grid8x8_p1.27mm", () => {
...
+  {
+    "number": undefined,
+    "x": "1.00mm",
+    "y": "-2.00mm",
+  },
+  {
+    "number": undefined,
+    "x": "2.00mm",
+    "y": "-2.00mm",
+  },
```
seveibar, line 107: is it supposed to not be centered by default? I feel like this is introducing a bug?

seveibar: what was the behavior before your change?

Rishikesh63: > what was the behavior before your change?

origin parameter affects only the grid's initial setup but not the final pad positioning.

Rishikesh63: > is it supposed to not be centered by default? I feel like this is introducing a bug?

yes

Rishikesh63:
@seveibar  review and gives feedback

tests/bga.test.ts
seveibar: So you're introducing a regression? Can you clarify?

Rishikesh63: No I am not introducing regression but  changing behavior by making the origin parameter actually affect pad placement.

seveibar (changes requested):
i don't think the x/y positions should change when you change the origin, just the numbering of the pins

the origin being at the top left means that the first pin is in the top left

Rishikesh63:
@seveibar  is it fine?

tests/bga.test.ts
```diff
@@ -35,10 +35,33 @@ test("bga64_w10_h10_grid8x8_p1.27mm", () => {
...
-  // 16pins, 4mm x 4mm, 8x8 grid, 1.27mm pitch
   const svgContent = convertCircuitJsonToPcbSvg(soup)
   expect(svgContent).toMatchSvgSnapshot(
     import.meta.path,
     "bga64_w10_h10_grid8x8_p1.27mm",
   )
 })
+
+test("bga footprint with top-left origin (default)", () => {
+  const soup = fp().bga(8).grid("3x3").p(1).soup()
```
imrishabh18, line 46:
```suggestion
  const circuitJsono = fp().bga(8).grid("3x3").p(1).soup()
```

Try to use the variable name `circuitJson` inplace of `soup`

Rishikesh63: in every file soup is used why i need to use circuitJson in place of soup

Rishikesh63: @imrishabh18  review now i replaced soup from circuitJson

Rishikesh63:
@imrishabh18  review now i have change soup to circuitjson

Rishikesh63:
@seveibar and @imrishabh18  please review

Issue #140, opened 2025-02-06
Implement SOT-89-3


Rishikesh63:
Here's are components on JLCPCB with this footprint: https://jlcsearch.tscircuit.com/components/list?package=SOT-89-3

There may be reference dimensions/designs on the [kicad viewer](https://tscircuit.github.io/kicad-viewer)

This issue may be eligible for a $10 bounty from the TSCircuit team (please review!)

PR #161, opened 2025-02-08
add sot89_3 footprint

techmannih:
fixes #140 
/claim #140

PR #162, opened 2025-02-08
Made a sot89-3 footprint

Rishikesh63:
/claim #140


Rishikesh63:
> @Rishikesh63 you have raised with duplicate code and logic of mine, if you want to claim need to change your logic and code
> 
> duplicate of #161

let it decided by reviewer many things are different in my code

Rishikesh63:
@imrishabh18  review my PR

imrishabh18 (approved):

Rishikesh63:
> can you add sot89_5 footprint with it?

why, it is sot89-3 issue?

Rishikesh63:
@seveibar  review my PR

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/footprinter",
   "type": "module",
-  "version": "0.0.115",
```
imrishabh18, line 4: We are not supposed to change the version!

src/fn/sot89.ts
```diff
@@ -0,0 +1,126 @@
...
+          "Invalid format. Expected 'sot89_N' where N is the number of pads.",
+      }),
+    width: z.string().default("4.80mm"),
+    height: z.string().default("4.8mm"),
+    padLength: z.string().default("1.44mm"),
+    padWidth: z.string().default("0.9mm"),
+    padGap: z.string().default("1.5mm"),
+  })
+  .transform((a) => {
+    const match = a.fn.match(/^sot89_(\d+)$/)
```
imrishabh18, line 23: What if the user passes the number other than 3 or 5? I think this will break in that case. We should allow only those number for which we have the footprint

imrishabh18 (changes requested):

src/fn/sot89.ts
Rishikesh63: now i only allowed sot89-3 and sot89-5 footprint

Rishikesh63:
@imrishabh18  review now

Rishikesh63:
@techmannih and @imrishabh18  review

Rishikesh63:
@techmannih  now i aligned it in a line

Rishikesh63:
@techmannih  does pitch should be as given KiCad viewer

Rishikesh63:
@techmannih  now i chaged the pad default length  and actual
image

Rishikesh63:
> @Rishikesh63 yeah, sot89-3 looks good

What about sot89-5

Anshgrover23 (commented):
@Rishikesh63 solved via #161

Issue #133, opened 2025-02-05
Implement TO-220


Rishikesh63:
Here's are components on JLCPCB with this footprint: https://jlcsearch.tscircuit.com/components/list?package=TO-220

There may be reference dimensions/designs on the [kicad viewer](https://tscircuit.github.io/kicad-viewer)

This issue may be eligible for a $10 bounty from the TSCircuit team (please review!)

PR #163, opened 2025-02-08
Made a TO-220 footprint

Rishikesh63:
/claim #133


Rishikesh63:
@imrishabh18  review my PR

tests/__snapshots__/to220.snap.svg
imrishabh18, line 1: This is `to220_2` cause it's having 2 `plated_holes`.

@seveibar What should be the naming convention for these, cause I can want the `to220_3` footprint as well?

tests/to220.test.ts
```diff
@@ -0,0 +1,9 @@
+import { test, expect } from "bun:test"
+import { convertCircuitJsonToPcbSvg } from "circuit-to-svg"
+import { fp } from "../src/footprinter"
+
+test("to220", () => {
+  const soup = fp.string("to220").circuitJson()
```
imrishabh18, line 6:
```suggestion
  const circuitJson = fp.string("to220").circuitJson()
```

imrishabh18 (changes requested):

tests/__snapshots__/to220.snap.svg
Rishikesh63: okay

Rishikesh63:
@imrishabh18  review now

src/fn/to220.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export type To220Def = z.input<typeof to220_def>
+
+export const to220 = (
+  raw_params: To220Def,
+): { circuitJson: AnySoupElement[]; parameters: any } => {
+  const parameters = to220_def.parse(raw_params)
+  const { fn, p, id, od, w, h } = parameters
+
+  const numPins = parseInt(fn.split("_")[1], 10) || 3
```
seveibar, line 33: do this parsing in a zod transform rather than in the fn, see other comment

src/fn/to220.ts
```diff
@@ -0,0 +1,111 @@
...
+    .refine((val) => /^to220_\d+$/.test(val), {
+      message:
+        "Invalid format. Expected 'to220_N' where N is the number of pins.",
+    }),
+  p: length.optional().default("5.0mm"),
+  id: length.optional().default("1.0mm"),
+  od: length.optional().default("1.9mm"),
+  w: length.optional().default("13mm"),
+  h: length.optional().default("7mm"),
+})
```
seveibar, line 23:
```suggestion
export const to220_def = z.object({
  fn: z
    .string()
    .default("to220_3")
    .refine((val) => /^to220_\d+$/.test(val), {
      message:
        "Invalid format. Expected 'to220_N' where N is the number of pins.",
    }),
  p: length.optional().default("5.0mm"),
  id: length.optional().default("1.0mm"),
  od: length.optional().default("1.9mm"),
  w: length.optional().default("13mm"),
  h: length.optional().default("7mm"),
}).transform(a => ({ ...a, numPins: a.fn.parse(...), fn: "to220" }))
```

seveibar:
CC @techmannih check out this implementation for t220_3

Rishikesh63:
> CC @techmannih check out this implementation for t220_3

also i am trying this for t89_3

Rishikesh63:
> > also i am trying this for t89_3
> 
> @Rishikesh63, yes, I have comments on sot89_3. Just to clarify, are you referring to t89_3 or sot89_3?

Sot89_3

Rishikesh63:
@seveibar  this line cause a error
}).transform(a => ({ ...a, numPins: a.fn.parse(...), fn: "to220" }))

# Unhandled error between tests
-------------------------------
23 | }).transform(a => ({ ...a, numPins: a.fn.parse(...), fn: "to220" }))
                                                       ^
error: Unexpected )
    at /mnt/c/Users/User/Desktop/footprinter/src/fn/to220.ts:23:51
    
    this error accurs because a.fn is a string not zod schema,in string we can't use parse,it's for zod schema

imrishabh18 (approved):

Rishikesh63:
> Hi @Rishikesh63, I noticed that the TO-220-3 (3 pads) looks different. Please double-check it. For TO-220-3 (3 pads), you seem to be using the same pitch as TO-220-2 (2 pads), but the TO-220-3 (3 pads) actually has a different pitch. Kindly verify and make the necessary adjustments.
> 
> @imrishabh18 what do you think?

i generalised it so pitch remain constant in both cases

seveibar (approved):
LGTM, it would be nice to also have the horizontal variants, feel free to create an issue to create the horizontal variants e.g. `to220_3_horizontal` and `to220_3_vertical`

nice work! /tip $2

seveibar:
@Rishikesh63 this seems to be broken #195, can you fix?

Rishikesh63:
> @Rishikesh63 this seems to be broken #195, can you fix?

Can I work on it on February 22nd since my exam will end on that day?

Issue #165, opened 2025-02-08
Allow creating footprints on footprinter website


seveibar:
![Image](https://github.com/user-attachments/assets/a3651656-f7c3-4972-beac-75e8148e8d59)

just a simple box that allows you to try a footprint

MustafaMulla29:
@ArnavK-09 Hey, can I work on this?

ArnavK-09:
> [@ArnavK-09](https://github.com/ArnavK-09) Hey, can I work on this?

already made pr #177

PR #177, opened 2025-02-09
feat: Allow creating footprints on footprinter website

ArnavK-09:
## outdated preview

image

image

/claim #165


package.json
```diff
@@ -12,7 +12,7 @@
     "dev": "cd previews && npm i && npm run storybook",
     "build": "tsup ./src/index.ts --format esm --dts --sourcemap",
     "format": "biome format . --write",
-    "generate-gallery": "bun run scripts/generate-gallery-page.ts"
```
seveibar, line 15: is there a script you forgot to delete? also `generate-data` has a banned word https://github.com/tscircuit/handbook/blob/main/guides/code.md

seveibar (commented):
can you import the tailwind cdn into the index.html file (DO NOT INSTALL tailwindcss, just add the CDN) and use regular tailwind classnames

Can you make the layout left/right for the textarea and preview (kind of doesn't look great imo)

also make sure you remove legacy files

package.json
ArnavK-09: I have modified script with new web method
And sorry for using data, lemme change that

ArnavK-09:
done side by side & responsive

image

Issue #167, opened 2025-02-08
Implement VSSOP-8-0.65mm


seveibar:
Here's are components on JLCPCB with this footprint: https://jlcsearch.tscircuit.com/components/list?package=VSSOP-8-0.65mm

There may be reference dimensions/designs on the [kicad viewer](https://tscircuit.github.io/kicad-viewer)

This issue may be eligible for a $10 bounty from the TSCircuit team (please review!)

MustafaMulla29:
@seveibar can you please review the pr [https://github.com/tscircuit/footprinter/pull/184](url)

MustafaMulla29:
@seveibar can you please review the pr https://github.com/tscircuit/footprinter/pull/184

PR #184, opened 2025-02-11
Implemented VSSOP-8-0.65mm

MustafaMulla29:
/claim #167


src/fn/vssop8.ts
```diff
@@ -0,0 +1,88 @@
...
+  pad_spacing: number
+}) => {
+  const { pn, pad_spacing } = parameters
+  const row = pn <= 4 ? -1 : 1 // Bottom row (1-4) or top row (5-8)
+  const col = ((pn - 1) % 4) - 1.5 // Distribute pads along width
+  return { x: col * pad_spacing, y: row * 0.5 * length.parse("3.0mm") }
+}
+
+// Generate pads for VSSOP-8
+export const vssop8Pads = (parameters: z.infer<typeof vssop8_def>) => {
```
seveibar, line 69: this is named like a variable, not a function, we should name things consistently- nouns for vars, verbs for functions
```suggestion
export const getVssop8Pads = (parameters: z.infer<typeof vssop8_def>) => {
```

src/fn/vssop8.ts
```diff
@@ -0,0 +1,88 @@
...
+    circuitJson: vssop8Pads(parameters).concat(
+      silkscreenLine as AnySoupElement,
+      silkscreenRefText as AnySoupElement,
+    ),
+    parameters,
+  }
+}
+
+// Get coordinates for VSSOP-8 pads
+export const getVssop8Coords = (parameters: {
```
seveibar, line 58: Notice, this should NOT be plural- a coordinate is a single x/y obj, coordinates would be an array

```suggestion
export const getVssop8PadCoord = (parameters: {
```

tests/vssop8.test.ts
```diff
@@ -0,0 +1,9 @@
+import { test, expect } from "bun:test"
+import { convertCircuitJsonToPcbSvg } from "circuit-to-svg"
+import { fp } from "../src/footprinter"
+
+test("vssop8", () => {
+  const circuitJson = fp.string("vssop8_0.65mm").circuitJson()
+  const svgContent = convertCircuitJsonToPcbSvg(circuitJson)
+  expect(svgContent).toMatchSvgSnapshot(import.meta.path, "vssop8_0.65mm")
```
seveibar, line 8: this snapshot doesn't look right, https://tscircuit.github.io/kicad-viewer/#Inductor_SMD.pretty/L_Sagami_CWR1277C.kicad_mod

seveibar (changes requested):
see comments

tests/vssop8.test.ts
MustafaMulla29: > this snapshot doesn't look right, https://tscircuit.github.io/kicad-viewer/#Inductor_SMD.pretty/L_Sagami_CWR1277C.kicad_mod

The issue was to create VSSOP-8, but I received this footprint instead. Could you clarify what needs to be adjusted?

MustafaMulla29:
@seveibar I have resolved the confilicts and tried to implement it again. Please review it!

MustafaMulla29:
> @MustafaMulla29 looks different! image


@techmannih ohh! Sorry, I was confused, I saw the implementation of the tssop8 on the footprinter and on kicad. Both were different so I implemented with the same method as of tssop8. 
So now I have to do exactly how it on kicad viewer? With only two lines above?

MustafaMulla29:
image

@techmannih is this how it should be?

MustafaMulla29:
> > @MustafaMulla29 looks different! image
> 
> @MustafaMulla29 we need exactly similar to this

Ok I'll make the changes and commit.

MustafaMulla29:
@techmannih you may review it now.

MustafaMulla29:
@seveibar can you review this

MustafaMulla29:
@techmannih Thanks for your response, I'll do that.

MustafaMulla29:
@techmannih Done!

MustafaMulla29:
@techmannih Resolved conflicts!

MustafaMulla29:
> sorry, form where you get this w and h, I am not able to see image

From kicad viewer

MustafaMulla29:
https://github.com/user-attachments/assets/579d03c5-8e61-410d-bad7-b2e8fc92ac9a

link: [https://tscircuit.github.io/kicad-viewer/#Package_SO.pretty/VSSOP-8_3x3mm_P0.65mm.kicad_mod](url)

@techmannih

MustafaMulla29:
width:
image

Height:
image

@techmannih

MustafaMulla29:
> @MustafaMulla29 correct which is 3.06 not 3.6mm, and measure h from top to bottom of silkscreen like this image

Ok I'll change that

seveibar (approved):
can you add a small pin1 marking.

I would recommend @techmannih does a quick review

MustafaMulla29:
image


@techmannih I tried my best adding the pin marker to the pin1, is this correct?

MustafaMulla29:
@techmannih please review, I added the pin marking and reduced the width(w) as well.

MustafaMulla29:
@techmannih please review, I added the pin marking and adjusted the width(w) and height(h).

ShiboSoftwareDev (changes requested):
please create more tests using different variants of these properties, untested properties caused many issues in the past
image

ShiboSoftwareDev:
@seveibar this should be enforced don't you think?

MustafaMulla29:
@ShiboSoftwareDev I have added more tests with different variants, please review.

MustafaMulla29:
@ShiboSoftwareDev my bad, do i have to get different variants from kicad viewer? I did the random ones. Please let me know, I'll make changes in the tests

ShiboSoftwareDev (approved):
@MustafaMulla29 lgtm just resolve the conflict

PR #200, opened 2025-02-17
improve to92 inline footprint

### LINKED ISSUE #196: to92_inline doesn't work
/bounty $8

Image

techmannih:
fix #196 
/claim #196


src/fn/to92.ts
```diff
@@ -28,10 +28,7 @@ export const to92_def = z.object({
   od: length.optional().default(".95mm"),
   w: length.optional().default("4.5mm"),
   h: length.optional().default("4.5mm"),
-  arrangement: z
-    .enum(["triangular", "inline"])
-    .optional()
-    .default("triangular"),
+  inline: z.boolean().optional().default(false),
```
imrishabh18, line 31:
```
  inline: z.boolean().optional().default(false),
```

What is the reason of renaming the variable?

seveibar (approved):

PR #204, opened 2025-02-18
Added male and female flag to pinrow footprint

AnasSarkiz:
/claim #202


src/fn/pinrow.ts
```diff
@@ -9,6 +9,12 @@ export const pinrow_def = z.object({
   p: length.default("0.1in").describe("pitch"),
   id: length.default("1.0mm").describe("inner diameter"),
   od: length.default("1.5mm").describe("outer diameter"),
+  male: z.boolean().optional().default(true).describe("for male pin headers"),
```
seveibar, line 12: you have to use a transform to set `male` to true only if `female` isn't defined
```suggestion
  male: z.boolean().optional().describe("for male pin headers"),
```

tests/pinrow.test.ts
```diff
@@ -5,5 +5,14 @@ import { fp } from "../src/footprinter"
...
   const soup = fp.string("pinrow5").circuitJson()
   const svgContent = convertCircuitJsonToPcbSvg(soup)
+  const pinrowJson = fp.string("pinrow5_female").json()
+  expect(pinrowJson).toMatchObject({
+    fn: "pinrow",
+    num_pins: 5,
+    p: 2.54,
+    id: 1,
+    od: 1.5,
+    female: true,
```
seveibar, line 15: this would fail, a bit sus you didn't put this in actually 😉 

```suggestion
    male: false
```

src/fn/pinrow.ts
```diff
@@ -3,13 +3,23 @@ import { length, type AnySoupElement } from "circuit-json"
...
-  od: length.default("1.5mm").describe("outer diameter"),
-})
+export const pinrow_def = z
+  .object({
+    fn: z.string(),
+    num_pins: z.number().optional().default(6),
+    p: length.default("0.1in").describe("pitch"),
+    id: length.default("1.0mm").describe("inner diameter"),
+    od: length.default("1.5mm").describe("outer diameter"),
+    female: z
```
seveibar, line 13: You should still allow male, the right pattern is to have _no defaults_ and let the transform handle it. male is still allowed here!!!

seveibar (approved):
See comment

src/fn/pinrow.ts
AnasSarkiz:
```tsx
 female: z.boolean().optional().describe("for female pin headers"),
  })
  .transform((data) => ({
    ...data,
    male: data.male ?? (data.female ? false : true), // Auto-set male if not explicitly provided
    female: data.female ?? false, // Default female to false if not provided
  }))
 ```

AnasSarkiz:
@seveibar  added error message if both male and female are true

```tsx
export const pinrow_def = z
  .object({
    fn: z.string(),
    num_pins: z.number().optional().default(6),
    p: length.default("0.1in").describe("pitch"),
    id: length.default("1.0mm").describe("inner diameter"),
    od: length.default("1.5mm").describe("outer diameter"),
    male: z.boolean().optional().describe("for male pin headers"),
    female: z.boolean().optional().describe("for female pin headers"),
  })
  .transform((data) => ({
    ...data,
    male: data.male ?? (data.female ? false : true), // Auto-set male if not explicitly provided
    female: data.female ?? false, // Default female to false if not provided
  }))
  .superRefine((data, ctx) => {
    if (data.male && data.female) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom,
        message:
          "'male' and 'female' cannot both be true it should be male or female.", //Error message if male and female are both true
        path: ["male", "female"],
      })
    }
  })

AnasSarkiz:
@seveibar can u check the latest commit

seveibar (approved):

Issue #203, opened 2025-02-18
Stampreceiver pins are marked incorrectly


seveibar:
pin 1 is always in the top right

/bounty $5

![Image](https://github.com/user-attachments/assets/7beb6125-298b-4827-8106-345775dde8e7)

Anshgrover23:
@techmannih it is not the way we work? you cannot assign yourself a issue until its assigned to you by @seveibar .

PR #206, opened 2025-02-19
Fixed stampreceiver pins are marked incorrectly

MustafaMulla29:
/claim #203


MustafaMulla29:
> @MustafaMulla29 I think we need pin1 on left and right as similar to other @seveibar what do you think?

The issue says pin1 should start from right side

imrishabh18 (changes requested):
I don't think this is the correct logic, you are adding the first pin on the right side rather it should be in the top left and it goes in an anticlockwise direction.

I can also see both the left and right are having top-to-bottom, which is not correct. The right should have bottom-to-top


cc: @seveibar

MustafaMulla29:
@imrishabh18 can you please clarify the things with me as the issue says "pin 1 is always in the top right" and you are saying its on left side.

imrishabh18:
Ohh okay, I think Seve might have typo-ed. The arrow points to the top left

MustafaMulla29:
image

@imrishabh18 is this fine? I checked with numbering(I'll remove it before pushing), as it is processing the top pads first and then going to the left pads, that's why the top right pad is getting number 2.

MustafaMulla29:
image

@imrishabh18  I checked with numberings(surely remove it before pushing), as it is processing the top pads first and then going to left side, that's the reason you see "2" on the top right pad.

is this fine?

MustafaMulla29:
image

@imrishabh18 is this good?(I'll remove the numbering before pushing)

imrishabh18:
@MustafaMulla29 Start from this position

image

MustafaMulla29:
@imrishabh18 Done!

imrishabh18 (approved):

MustafaMulla29:
@seveibar Can you merge this pr

seveibar:
@techmannih can you help review this and make sure it fixes the original issue?

MustafaMulla29:
@techmannih added pin numbers, please review

seveibar (commented):
the pin names are inside the pads rather than outside the pads (this board won't be manufactured if there is text on the pads)

MustafaMulla29:
> the pin names are inside the pads rather than outside the pads (this board won't be manufactured if there is text on the pads)

I'll quickly change it

MustafaMulla29:
@seveibar should I remove pin numbers?

seveibar (approved):

PR #216, opened 2025-02-22
add rows param in pinrow footprint

techmannih:
fixes #https://github.com/tscircuit/core/issues/432

/claim #https://github.com/tscircuit/core/issues/432


tests/__snapshots__/pinrow8_rows4.snap.svg
imrishabh18, line 1: Shouldn't the `REF` be at the top left?

Issue #198, opened 2025-02-17
sot223 doesn't work


seveibar:
![Image](https://github.com/user-attachments/assets/1e58117d-e3a1-4823-ba4f-570529bf3df1)

/bounty $5

Anshgrover23:
@seveibar we do not have support of sot223

PR #217, opened 2025-02-22
 Fix: Resolved sot223 issue #198

deekshatomer:
/claim  #198

This PR fixes issue #198 where the sot223 footprint was not working correctly
Changes Made
- Updated the sot223 footprint logic.
- Verified that `fpInstance.sot223(6)` now returns the expected parameters.

### Testing
- Ran the relevant tests to confirm the fix.
- Verified that the output `{ sot223: true, fn: 'sot223', num_pins: 6 }` is correct.

### Notes
- Let me know if any additional changes are needed.


imrishabh18:
Wrong, I don't even know what this is supposed to do!!

PR #218, opened 2025-02-22
update snapshot due to bound issue

techmannih:
update snapshot due to bound issue


package.json
```diff
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "@tscircuit/mm": "^0.0.8",
+    "circuit-json-to-connectivity-map": "^0.0.18",
```
seveibar, line 38: dev dependency!!!!!!

seveibar (changes requested):
dev dep

seveibar (approved):

PR #233, opened 2025-03-02
remove unwanted anchor text with footprint

techmannih:
Ref #https://github.com/tscircuit/core/issues/673

### DIFF
--- a/src/helpers/passive-fn.ts
+++ b/src/helpers/passive-fn.ts
@@ -200,12 +200,12 @@ export const passive = (params: PassiveDef): AnySoupElement[] => {
       silkscreenLine,
       silkscreenRefText,
     ]
-  } else {
-    return [
-      rectpad(["1", "left"], -p / 2, 0, pw, ph),
-      rectpad(["2", "right"], p / 2, 0, pw, ph),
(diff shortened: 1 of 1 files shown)


imrishabh18 (commented):
LGTM! But I don't know if this the right way to fix it. Cause footprint resolving schematic issue 🤷‍♂️. I guess we could have not added the anchor_text in the `circuit_to_svg` schematic or maybe this is the right way. I am not sure TBH.

@seveibar take a look

seveibar (changes requested):
not the right fix

seveibar:
it's a core issue, when we output the schematic net labels we want to select the most appropriate ones

seveibar:
RE: sizing, we also want to properly size schematic components in core

techmannih:
https://github.com/user-attachments/assets/21f14610-2945-4e7e-b038-948dba0e5740

techmannih:
@seveibar @imrishabh18 I am tried this with two times, please you can review in attachment.

techmannih:
> RE: sizing, we also want to properly size schematic components in core

Yeah sizing will fix in core

PR #253, opened 2025-03-24
add BGA origin parameter support

### LINKED ISSUE #107: bga origin parameter has no effect

Anshgrover23:
/claim #107
Fixes #107


seveibar (commented):
Snapshot tests?

seveibar (changes requested):
Clearly broken from the snapshots, the ref position is changing

seveibar (changes requested):
pin1 isn't marked in the snapshot, so somebody assembling with this footprint wouldn't know the correct orientation (and I can't tell which pin is actually pin1)

seveibar (approved):

PR #255, opened 2025-05-07
introduced jumper footprint

ShiboSoftwareDev:
there is no way to know the pins and connect them using source traces, so I used pcb traces


seveibar (commented):
this is OK but the smtpad size is super arbitrary, most jumpers i would assume are plated holes. I might call this `jumperpads` and have the standard `pw` (pad width) and `ph` (pad height) as parameters

looking good though

src/fn/jumper.ts
```diff
@@ -0,0 +1,72 @@
...
+              },
+              {
+                x: (to - 1) * padSpacing,
+                y: 0,
+                width: 0.5,
+                layer: "top",
+                route_type: "wire",
+              },
+            ],
+          })
```
seveibar, line 57: yea i think this is fine, good call :thumbs-up:

ShiboSoftwareDev:
@seveibar can I call it solderjumper?
image

seveibar:
@ShiboSoftwareDev hmm `solderjumper_bridged` and `solderjumper_bridged(1,2)`

seveibar:
@ShiboSoftwareDev sgtm!

ShiboSoftwareDev:
@seveibar solderjumper_bridged23 and solderjumper_bridged123 sound better

PR #256, opened 2025-05-08
Feat: each pin in pinrow ft has silkscreenRef

Abse2001:


seveibar (changes requested):
Great work! The text should be {PIN1}, {PIN2} etc. for replacement in core

for a future pr: we could also add a flag to indicate the text position. The api will be a bit hard to come up with so lets do in future.

seveibar (approved):
I dont think its your fault, but the ordering on a bunch of these pins is wrong, pin1 is always the top left and it should go counter clockwise to count up cc @techmannih maybe can fix

techmannih:
Yeah seve I will take a look on it.

PR #257, opened 2025-05-09
fix: ordering for pinrows

techmannih:


src/fn/pinrow.ts
```diff
@@ -79,11 +79,7 @@ export const pinrow = (
   }
 
   // Compute group reference position centered horizontally
-  const perRow = rows > 1 ? Math.ceil(numPins / rows) : numPins
-  const xStartGroup = -((perRow - 1) / 2) * p
-  const groupTextX = xStartGroup + ((perRow - 1) / 2) * p
-  const groupTextY = rows * p
-  const refText: SilkscreenRef = silkscreenRef(groupTextX, groupTextY, 0.5)
+  const refText: SilkscreenRef = silkscreenRef(0, p, 0.5)
```
imrishabh18, line 82: Why the groupTextX is 0 ?

seveibar (changes requested):
Pin numbers should go counter clockwise

seveibar (changes requested):
Pin number 1 is at the top left and always go counter clockwise

seveibar (approved):

PR #259, opened 2025-05-09
solderjumper now inserts source_trace

ShiboSoftwareDev:
yalc tested on core


seveibar (changes requested):
Ehh this feels pretty weird- i think source_* should be handled in core and footprinter should handle the pcb_* stuff. Core has better access to the port ids and can do it properly without pseudo-id insertions

ShiboSoftwareDev:
@seveibar I want to create the trace in core, but how do I send the connected ports without sending a source trace element?
I think inserting random pcb_traces is not the as you would connect two components that are not in the same connectivity net

ShiboSoftwareDev:
this is how it's handled in core:
image

seveibar:
@ShiboSoftwareDev i see what you're saying, and its ok to create a source trace in core to represent the trace, but i dont think footprinter should output source traces because it's not reading the source. It should also output pretty exact positions and a source traces dont have a position

This will allow us to do more custom pcb trace configurations in the future

PR #260, opened 2025-05-10
refactor jumper for better connections

ShiboSoftwareDev:
it had fixed trace width which is problematic


seveibar (approved):
Ok for now but traceWidth may need to be parameter in future, sometimes tiny trace wodths are very desirable since these are made to be cut

PR #263, opened 2025-05-10
added start/end_pcb_port_id to help with connectivity map

ShiboSoftwareDev:
this way we can know which port/pin to add

we already know what component it is in core

### DIFF
--- a/src/fn/solderjumper.ts
+++ b/src/fn/solderjumper.ts
@@ -60,7 +60,7 @@ export const solderjumper = (params: {
             pcb_trace_id: "",
             route: [
               {
-                start_pcb_port_id: `pin${from}`,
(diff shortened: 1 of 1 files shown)


seveibar (approved):
Can you prefix the id so that we know its a pseudo id? Ie “footprinter_pseudo_pin{to}”, otherwise it will be really confusing if ids are not properly replaced. Alternatively use the existing template syntax used by footprinter “{PIN1}” ie “{PIN${to}}”

seveibar:
@ShiboSoftwareDev to be clear, “{PIN1}” is the syntax i was proposing to make it obvious it is a template, following the kicad {REF} and {VAL} convention

seveibar (approved):

PR #268, opened 2025-05-20
Add squareplating support in pinrow

### LINKED ISSUE #267: add support squareplating in pinrows

techmannih:
fixes #267


seveibar (changes requested):
- be default, always accept square plating. No need to handle the case where the first pin isn't square plated for now
- Derive the rectangular pad width and height from the provided outer diameter

No new parameters needed

PR #271, opened 2025-05-21
new properties for pinrow: labelrotation adn labelposition

ShiboSoftwareDev:


src/fn/pinrow.ts
```diff
@@ -19,9 +19,15 @@ export const pinrow_def = z
     od: length.default("1.5mm").describe("outer diameter"),
     male: z.boolean().optional().describe("for male pin headers"),
     female: z.boolean().optional().describe("for female pin headers"),
+    labelrotation: z.enum(["0", "90", "180", "270"]).optional().default("0"),
+    labelposition: z
+      .enum(["Up", "Down", "Left", "Right"])
```
seveibar, line 24: Lowercase

ShiboSoftwareDev: @seveibar I made it like this to be able to parse it, do you have a better method of parsing?

ShiboSoftwareDev: instead of "_labelpositionUp" "labelposition(up)" ?

seveibar: Im really confused. What is a label? Can you use either pin label or refdes? There are pin labels and the refdes label. I think youre talking about the refdes position is that right?

if so, it would be a nine point anchor

src/fn/pinrow.ts
```diff
@@ -19,9 +19,15 @@ export const pinrow_def = z
     od: length.default("1.5mm").describe("outer diameter"),
     male: z.boolean().optional().describe("for male pin headers"),
     female: z.boolean().optional().describe("for female pin headers"),
+    labelrotation: z.enum(["0", "90", "180", "270"]).optional().default("0"),
```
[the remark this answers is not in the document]
seveibar: *parallelpinlabels

src/fn/pinrow.ts
seveibar: I dont think this propery makes sense, that said labelpositionup would be fine, flags are fine, but this label isnt named well enough that ai will understand what it foes rn

seveibar (changes requested):
It's really hard for me to understand what's going on because the snapshot names are wrong (they don't match the footprint), so it's hard to put together what the actual footprint -> svg is

seveibar:
image

looks like a tiny bit of text alignment

PR #277, opened 2025-05-25
add optional 'pw' and 'pl' to extendSoicDef

techmannih:
need to make PW and PL in  extendSoicDef optional for another footprint


seveibar (commented):
can you add a simple test for this?

seveibar (approved):

PR #283, opened 2025-05-30
fixed pinrow pinlabel orientation options

ShiboSoftwareDev:
it's perfect now


seveibar (commented):
Can we remove pinlabelright, left etc? Will do full review in sec

seveibar (approved):
Cant do full review but api is correct! Will do full review in 20 mins or so, feel free to merge bc api alone is substantially better

ShiboSoftwareDev:
oops

PR #288, opened 2025-06-07
add support polygon smtpad

techmannih:


seveibar (changes requested):
no test and unclear when this would be used, how does the user provide points etc.

seveibar:
To me this makes no sense, the user won't provide a polygon via a string

PR #289, opened 2025-06-07
add fuse footprint

techmannih:


seveibar:
fuse is not a standard footprint

PR #295, opened 2025-06-13
fix: correct pad dimension logic in passive function

techmannih:
always using same pw and ph?
image


seveibar (approved):
fix the PR title!!

Issue #315, opened 2025-07-07
0603 footprint doesn't look correct


ShiboSoftwareDev:
first, verify if it's accurate
ours:
![Image](https://github.com/user-attachments/assets/ebfb22f7-0d98-4ce4-8235-c48ad94eddc0)
reference:
![Image](https://github.com/user-attachments/assets/8bd3f73f-0559-44cb-9886-1dfcce2647fb)

ShiboSoftwareDev:
@MustafaMulla29 can you try to fix it? if not I'll check it out

MustafaMulla29:
> [@MustafaMulla29](https://github.com/MustafaMulla29) can you try to fix it? if not I'll check it out

Ok I'll try it

PR #316, opened 2025-07-07
Fixed the 0603 footprint dimensions

MustafaMulla29:
/fixes https://github.com/tscircuit/footprinter/issues/315


seveibar (approved):
you do need to update the snapshots for other tests that use 0603

CC @techmannih please review

PR #320, opened 2025-07-10
Introduce JST-SH connector variant

techmannih:


src/fn/jst.ts
```diff
@@ -4,58 +4,139 @@ import {
...
-  w: length.optional().default("6mm"),
-  h: length.optional().default("5mm"),
+  p: length.optional(),
+  id: length.optional(),
+  pw: length.optional(),
+  pl: length.optional(),
+  w: length.optional(),
+  h: length.optional(),
+  sh: z.boolean().optional(),
+  ph: z.boolean().optional(),
```
seveibar, line 21: add a `.describe` here, i don't know what this is, same for sh

PR #322, opened 2025-07-14
Introduce JST-SH Connector footprint variant

techmannih:


ShiboSoftwareDev (approved):
looks good, bug can I see the reference for the dimensions?

src/fn/jst.ts
```diff
@@ -4,58 +4,171 @@ import {
...
-  pl: length.optional().default("1.20mm"),
-  w: length.optional().default("6mm"),
-  h: length.optional().default("5mm"),
+  p: length.optional(),
+  id: length.optional(),
+  pw: length.optional(),
+  pl: length.optional(),
+  w: length.optional(),
+  h: length.optional(),
+  sh: z
```
seveibar, line 20: these need to have .describe!!!!!

particularly sh, and ph, but honestly ALL of them should have .describe!!!

PR #338, opened 2025-07-31
feature: add compareFootprinterVsKicad to validate footprint accuracy against KiCad

Abse2001:


seveibar (commented):
This is cool, can you show the kicad and footprinter footprints side by side by using transformPcbElements from circuit json util?

seveibar (changes requested):
Sorry feeling better today so can do a bit better for the review:

* in the snapshot, show the exact name of the footprinter string and the exact name for the kicad footprint, not KICAD_FP and LOCAL_FP
* one test per file. If it is a kicad parity test, put it in tests/kicad-parity
* do not change any existing footprints in this PR
* In the snapshot, add silkscreen text at the bottom than shows the pad difference as a %

seveibar (changes requested):
sorry i'm just coming to after having a pretty bad fever (I'm still a bit light-headed etc. but able to read this through a bit more now)

We should be able to compare ANY TWO footprinter and kicad footprints, this code is specific to resistors.

tests/fixtures/padDifference.ts
```diff
@@ -0,0 +1,264 @@
...
+}
+
+export async function padDifference(
+  imperialOrMetric: string,
+  footprintName: string,
+): Promise<{
+  avgRelDiff: number
+  combinedFootprintElements: any[]
+}> {
+  const normalized = imperialOrMetric.replace(/_(imp|metric)$/, "")
```
seveibar, line 37: None of this logic should be here, you should just render the footprint to circuit json, then compute what area is covered by smtpads

```suggestion
```

tests/fixtures/compareFootprinterVsKicad.ts
```diff
@@ -0,0 +1,153 @@
...
+  type: "pcb_smtpad"
+  x: number
+  y: number
+  width: number
+  height: number
+  port_hints: string[]
+}
+
+export async function compareFootprinterVsKicad(
+  imperialOrMetric: string,
```
seveibar, line 17:
```suggestion
  footprinterString: string,
```

tests/fixtures/compareFootprinterVsKicad.ts
```diff
@@ -0,0 +1,153 @@
...
+  footprintName: string,
+): Promise<{
+  avgRelDiff: number
+  combinedFootprintElements: any[]
+}> {
+  let size = imperialOrMetric.endsWith("_imp")
+    ? footprintSizes.find((s) => s.imperial === imperialOrMetric)
+    : footprintSizes.find((s) => s.metric === imperialOrMetric) ||
+      footprintSizes.find((s) => s.imperial === imperialOrMetric)
```
seveibar, line 26: remove this

tests/fixtures/compareFootprinterVsKicad.ts
```diff
@@ -0,0 +1,153 @@
...
+  if (!res.ok)
+    throw new Error(`Failed to fetch ${normalizedFootprintName}: ${res.status}`)
+
+  const data = (await res.json()) as any[]
+
+  const kicadPads = data.filter(
+    (e) =>
+      e.type === "pcb_smtpad" &&
+      Array.isArray(e.port_hints) &&
+      (e.port_hints.includes("1") || e.port_hints.includes("2")),
```
seveibar, line 47:
```suggestion
```

tests/fixtures/compareFootprinterVsKicad.ts
```diff
@@ -0,0 +1,153 @@
...
+    0,
+  )
+
+  const avgRelDiff = Math.abs(refArea - kicadArea) / (refArea + kicadArea) || 0
+  const diffPercent = avgRelDiff * 100
+
+  // Visual placement
+  const transformedReference = transformPcbElements(
+    circuitJson,
+    translate(0, 0),
```
seveibar, line 71: why transform if you're not moving anything?

seveibar (commented):
general technique is looking good!

PR #346, opened 2025-08-09
Fix DIP pin row spacing and adjust silkscreen to match hole centers and clearance

Abse2001:


seveibar (commented):
@Abse2001 you MUST provide the kicad snapshots, this is a major breaking change so you must prove correctness, where are the snapshots?

PR #348, opened 2025-08-09
Fix DIP pin row spacing and adjust silkscreen to match hole centers and clearance

Abse2001:


seveibar (commented):
looks good but there's an issue with the fabrication text

seveibar:
@Abse2001 👍 awesome work on this, quality is going to be much better/more reliable

seveibar (approved):

PR #349, opened 2025-08-10
Fix silkscreen pin text alignment to respect anchor placement

techmannih:


seveibar (changes requested):
No idea what this fixes

PR #359, opened 2025-08-11
using the exact flags as kicad for parity testing

Abse2001:


ShiboSoftwareDev:
diffs are huge, are you after this going to change the numbers according to the spec?

Abse2001:
@ShiboSoftwareDev if I found the specs lol yeah ofc

PR #369, opened 2025-09-27
update snapshots

MustafaMulla29:


imrishabh18 (commented):
How are these snapshots updated? Where is the change? these are just svg's.

Strange why didn't the test workflow fail

MustafaMulla29:
> How are these snapshots updated? Where is the change? these are just svg's.
> 
> Strange why didn't the test workflow fail

Yea i was checking the main branch running the bun test and the snapshots changed without any change

Issue #265, opened 2025-05-13
ability to disable component ref schematic text


ShiboSoftwareDev:
![Image](https://github.com/user-attachments/assets/5ba6e819-e650-4965-94af-0354b8d60076)

PR #389, opened 2025-10-22
fix: add disable option for bga ref label

joncherry:
related to #265 

This PR is a smaller PR for just BGA to validate the approach to solving #265 after this is merged I will make a larger PR to use the same approach on all the other footprints which will close the issue.

### Summary
the labels in the footprint snapshots such as "{ref}" originate from `src/helpers/silkscreenRef.ts` where "{ref}" is assigned on the circuit-json text field. We can't just disable the text field because it is in the circuit-json repo https://github.com/tscircuit/circuit-json/blob/main/src/pcb/pcb_silkscreen_text.ts#L58 but we can set it to empty string.

### Side Note
this seems very similar to the name field for footprints in the tscircuit pcb view. You can also just set those to empty string and it usually looks okay
```ts
export default () => (
<board width="10mm" height="10mm">
  <resistor
    name=""
    resistance="10k"
    footprint="kicad:Resistor_SMD/R_0402_1005Metric"
    pcbX={0}
  />
</board>
)
```
if you remove the name field the label in the pcb view becomes "unnamed_resistor"

### Results
Screenshot from 2025-10-22 13-13-04
Screenshot from 2025-10-22 13-13-54

I would add a video but it would not show anything valuable that isn't shown by those screenshots :point_up:


src/fn/bga.ts
```diff
@@ -12,6 +13,7 @@ import { type PcbSilkscreenPath } from "circuit-json"
 
 export const bga_def = z
   .object({
+    ...base_def,
     fn: z.string(),
     num_pins: z.number().optional().default(64),
```
techmannih: I think you do not add new file for noref flag, you can add new flag "noref" as same as other

techmannih: Wait a sec — we can probably reuse this base_def with other footprints.
If that makes sense, then there’s no need to add it here. It would be better to keep it in a separate file, as it is now.
We can confirm this approach with @seveibar before finalizing.

ShiboSoftwareDev (approved):
looking good

seveibar (changes requested):
`base_def.extend` is the correct way to have extensible zod types

src/fn/bga.ts
```diff
@@ -166,6 +168,9 @@ export const bga = (
     (grid.y * p) / 2,
     0.2,
   )
+  if (parameters.noref) {
+    silkscreenRefText.text = ""
+  }
```
seveibar: empty strings are ALWAYS an anti-pattern! not allowed!

`noref` should be implemented in a post-processing transform as discussed in discord

src/helpers/postProcessing.ts
```diff
@@ -0,0 +1,12 @@
+import type { AnyCircuitElement } from "circuit-json"
+import { applyNoRef } from "./apply-noref"
+import { applyOrigin } from "./apply-origin"
+
+export const postProcessing = (
```
seveibar: really bad function name, i don't think this refactor makes sense

src/footprinter.ts
```diff
@@ -309,7 +309,7 @@ export const footprinter = (): Footprinter & {
           if ("fn" in target && FOOTPRINT_FN[target.fn]) {
             return () => {
               const { circuitJson } = FOOTPRINT_FN[target.fn](target)
-              return applyOrigin(circuitJson, target.origin)
+              return postProcessing(circuitJson, target)
```
seveibar: just do `applyOrigin(applyNoRef(` here

seveibar (commented):
generally ok but postProcess is a really vague name, not acceptable

src/footprinter.ts
joncherry: I was trying to do it like a piece of new code architecture, where multiple post-processors could be applied to the circuitJson. You're right it is probably too early for that. probably need at least 3 or 4 post processors before creating that architecture.

I have never been fond of that name choice. I completely agree that it would need to be something else that I have yet to think of.

seveibar: yea the issue is also the name is vague, it could be more acceptable if it was named something like `applyAttributeModifications` but even that is too vague- we need more apply functions in order to understand what is really good name, it's possible that it's more clear to just have a `compose(`-style function, but more apply* functions are needed first i think

joncherry: fixed.

PR #390, opened 2025-10-25
add support tqfp

Aqil-Ahmad:
fixes: tscircuit/tscircuit#1172
/claim tscircuit/tscircuit#1172

added support for tqfp


Aqil-Ahmad:
> ah @Aqil-Ahmad this is quite disappoint, tqfp is not qfp and a simple google search shows that

uhhh really sorry for that! i mapped tqfp to qft assuming they were interchangeable when they're not.

PR #399, opened 2025-11-01
Fix lqfp footprint

rushabhcodes:
## Before:
image

## After: 
image


seveibar (approved):

techmannih:
> @techmannih what is the right way to this

No issue minor issue. I will be fix it,

rushabhcodes:
@techmannih  depending on the pitch the pl also changes for different varients, if you do implement it, do tag me

PR #405, opened 2025-11-04
Fix sot323 footprint

nailoo:


ShiboSoftwareDev:
how did the kicad footprint change?

seveibar (approved):

PR #408, opened 2025-11-07
Remove sot25 component and associated tests

rushabhcodes:
This pull request removes all code and references related to the `sot25` footprint type from the codebase. The primary changes involve cleaning up exports, type definitions, implementation files, and related tests.

`sot25` is unnecessary as it is already covered in `sot23` as `sot23_5` 

### Removal of `sot25` footprint

* Removed the export of `sot25` from the `src/fn/index.ts` file, so it is no longer available for import elsewhere in the project.
* Deleted the entire implementation of the `sot25` footprint from `src/fn/sot25.ts`, including its type definition and logic.
* Removed the `sot25` entry from the `Footprinter` type in `src/footprinter.ts`, eliminating its type support.

### Removal of related tests

* Deleted the test file `tests/sot25.test.ts`, which previously checked the `sot25` footprint functionality.

### DIFF
--- a/.github/workflows/bun-test.yml
+++ b/.github/workflows/bun-test.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup bun
         uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          bun-version: 1.3.1
 
       - name: Install dependencies
         run: bun install


rushabhcodes:
> we are not using this one?

no, `sot23_5` is used even in jscad

PR #422, opened 2025-11-16
add a footprinter flag for pinrow to allow the pins to faceup

nailoo:


seveibar (changes requested):
read the conversation carefully, the footprint does not change

rushabhcodes (changes requested):
This is not required

Issue #180, opened 2025-02-10
Implement SPDIP-28


Abse2001:
Here's are components on JLCPCB with this footprint: https://jlcsearch.tscircuit.com/components/list?package=SPDIP-28

There may be reference dimensions/designs on the [kicad viewer](https://tscircuit.github.io/kicad-viewer)

This issue may be eligible for a $10 bounty from the TSCircuit team (please review!)

bitdeckai:
I can take this. Plan: add SPDIP-28 footprint parameters aligned to KiCad/JLC reference, wire it into the footprinter string mapping, and add a snapshot test to validate output.

bitdeckai:
Opened PR: https://github.com/tscircuit/footprinter/pull/508

Golradreth:
/attempt #180

I am going to implement this as a narrow parser alias: normalize SPDIP package strings such as spdip28 and SPDIP-28_W7.62mm to the existing DIP generator, then add focused parser coverage plus a KiCad parity snapshot against DIP-28_W7.62mm so the previous snapshot/test gap is covered.

Joel8rub:
Implemented in PR #629: https://github.com/tscircuit/footprinter/pull/629

Verification:
- `npm run build` passed
- `npx bun test tests/spdip.test.ts` passed (2 tests)
- `npx biome check src/fn/spdip.ts tests/spdip.test.ts` passed

BossChaos:
🎯 Working on implementing this footprint! Will submit a PR with full implementation and tests. Wallet: 0xdaE5d307339074A24F579dB48e7c639359D94904

Heyzerohey:
/claim #180 via https://github.com/tscircuit/footprinter/pull/643

chengdazhi:
/attempt #180

PR: https://github.com/tscircuit/footprinter/pull/703

Implements `spdip` (SPDIP-28) as a narrow DIP layout (7.62mm row width, 2.54mm pitch), verified against KiCad `DIP-28_W7.62mm` with a new parity test (courtyard IoU 99.84%, pads fully overlap). Full `bun test` suite passes (437/437).

PR #426, opened 2025-11-17
fix: add footprint test for spdip-28

joncherry:
/closes #180

# Summary
It appears to me that the issue for SPDIP-28 fits into the dip function at `src/fn/dip.ts` so I have added a test for that.

# Evidence
I looked at this doc. When converting from inches to mm in the doc, the my new test seems to have good dimensions that match the doc. 0.300 inch width and 0.100 inch pitch become 7.62mm and 2.54mm respectively. The pin diameter of 0.018 inch becomes 0.457 mm which should fit into the default ID of the pin holes on the test which seems to be 1.0mm
https://ww1.microchip.com/downloads/aemDocuments/documents/corporate-responsibilty/environmental/material-compliance-documents/28L_SPDIP_300_SP_C04-070B.pdf

also the footprint visually resembles the footprint from jlc search:
https://jlcsearch.tscircuit.com/components/list?package=SPDIP-28
Screenshot from 2025-11-17 13-36-27

Here's the Google AI overview of what SPDIP-28 means:

<details>
<summary>expand</summary>
In a PCB context,
SPDIP-28 refers to a specific type of integrated circuit (IC) package, describing its physical characteristics and pin count. 
Meaning of each part

    S - Shrink or Skinny. This indicates a smaller form factor compared to a standard Plastic Dual In-line Package (PDIP), typically meaning:
        Narrower body width: Often 0.3 inches (300 mils) wide, whereas a standard high-pin-count DIP (like 28 pins) might typically be 0.6 inches wide.
        Denser pin pitch (less common): In some cases, it can refer to a package with a smaller pin-to-pin pitch, such as 0.07 inches (1.778 mm), instead of the standard 0.1 inches (2.54 mm) pitch. The 0.3-inch width with 0.1-inch pitch is the more common use for 28-pin devices from manufacturers like [Microchip](https://www.microchip.com/mymicrochip/filehandler.aspx?ddocname=en595838).
    PDIP - Plastic Dual In-line Package. This describes the type of construction (plastic body) and the arrangement of pins in two parallel rows along the longer sides of the rectangular package.
    28 - 28 pins. This specifies the total number of electrical connecting pins the component has. 

Key characteristics for PCB design

    Mounting: It is a through-hole package, meaning its pins are designed to be inserted into holes drilled in the PCB and soldered on the other side.
    Dimensions: The most crucial dimensions for PCB layout are the number of pins (28), the pin pitch (usually 0.1 inches/2.54 mm), and the row spacing (typically 0.3 inches/300 mils for "skinny" versions).
    Application: These packages were widely used for microcontrollers and other ICs, and they require a compatible PCB footprint or a DIP socket, especially in prototyping or older consumer electronics.
</details>


techmannih (commented):
@joncherry i think you need to implement spdip footprint cc @seveibar 
Screenshot_2025-11-22-08-48-44-66_40deb401b9ffe8e1df2f1cc5ba480b12

joncherry:
@techmannih see https://ww1.microchip.com/downloads/aemDocuments/documents/corporate-responsibilty/environmental/material-compliance-documents/28L_SPDIP_300_SP_C04-070B.pdf from my description comment above which is the pdf that I used as my source for dimensions.

Edit: looks like kicad viewer has smdip but doesn't have spdip.

joncherry:
> can you add kicad parity test too if possible

@rushabhcodes For SPDIP, there is no footprint in kicad viewer or kicad libraries https://gitlab.com/kicad/libraries/kicad-footprints.

However I did find SPDIP on these kicad repos for symbols.

https://gitlab.com/kicad/libraries/kicad-library-utils/-/blob/master/symbol-generators/avr_da_db/avr_da_db_generator.py#L78

https://gitlab.com/kicad/libraries/kicad-library-utils/-/blob/master/symbol-generators/avr_da_db/avr_da_db_generator.py#L86

and

https://gitlab.com/kicad/libraries/kicad-symbols/-/blob/master/MCU_Microchip_PIC18.kicad_sym#L2957

https://gitlab.com/kicad/libraries/kicad-symbols/-/blob/master/MCU_Microchip_PIC18.kicad_sym#L2976

Which seems to further support that the footprint for SPDIP is a dip footprint.

PR #429, opened 2025-11-21
fix: update SOT89 dimensions and snapshots for consistency

rushabhcodes:
This pull request makes a small change to the default width value in the `sot89_def` schema definition, updating it from "4.20mm" to "4.00mm".


rushabhcodes:
> is there a bug in the kicad parity test? why does it still match after we change the dimensions?

I changed the body width, the kicad test only compares pads length and width

rushabhcodes:
@seveibar please review

Issue #183, opened 2025-02-10
Implement SPDIP-28


Abse2001:
Here's are components on JLCPCB with this footprint: https://jlcsearch.tscircuit.com/components/list?package=UTDFN-4-EP(1x1)

There may be reference dimensions/designs on the [kicad viewer](https://tscircuit.github.io/kicad-viewer)

This issue may be eligible for a $10 bounty from the TSCircuit team (please review!)

BossChaos:
🎯 Working on implementing this footprint! Will submit a PR with full implementation and tests. Wallet: 0xdaE5d307339074A24F579dB48e7c639359D94904

Heyzerohey:
/claim #183 via https://github.com/tscircuit/footprinter/pull/644

thienanwspace:
/claim #183

PR: https://github.com/tscircuit/footprinter/pull/700

Implements UTDFN-4-EP(1x1) footprint with exposed pad, alias normalization, and tests.

chengdazhi:
/attempt #183

PR: https://github.com/tscircuit/footprinter/pull/704

Implements `utdfn` for UTDFN-4-EP (1x1mm): 4 chamfered signal pads + 0.53mm exposed pad, 0.65mm pitch, per the Diodes X2-DFN1010-4 (Type B) suggested pad layout (also compatible with the Micrel 4-pin 1x1mm Thin MLF "MT" package used by MIC5366-xYMT). Includes snapshot/alias tests and a pad-vs-exposed-pad overlap check. Full `bun test` suite passes (437/437).

JoTalbot:
Octopus automation reviewed this bounty as a potential contribution target. Issue: Implement UTDFN-4-EP(1x1). Reward: 10.0. I am preparing a scoped implementation and will avoid duplicate work.

PR #437, opened 2025-12-06
Add UTDFN-4-EP footprint and tests

Devesh36:
**Files**
- Implementation: `src/fn/utdfn4ep.ts`
- Tests: `tests/utdfn4ep.test.ts`
- Snapshots: `tests/__snapshots__/utdfn4ep_default.snap.svg` and `tests/__snapshots__/utdfn4ep_overrides.snap.svg`

**Summary**
- `utdfn4ep` generates a UTDFN-4-EP footprint sized for a 1.0 × 1.0 mm body.
- Produces four symmetric side pads (pins 1–4) and a central exposed pad (EP) with ID `5`.

default - 
Screenshot 2025-12-06 at 10 00 49 PM

override - 
Screenshot 2025-12-06 at 10 00 54 PM

/fix #183

### DIFF
--- a/src/fn/utdfn4ep.ts
+++ b/src/fn/utdfn4ep.ts
@@ -12,13 +12,12 @@ import { z } from "zod"
  * - https://ww1.microchip.com/downloads/en/DeviceDoc/MIC5366-1.8YMT-TZ-DS20005619G.pdf (page 11)
  */
 export const utdfn4ep_def = z.object({
- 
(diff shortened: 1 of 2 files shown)


techmannih (commented):
The KiCad parity test, reference text is missing in your footprint. This is not the workflow we follow for creating new footprints, and the required flag is also missing in footprinter function.

Issue #396, opened 2025-10-30
Implement SOT-343-4


rushabhcodes:
Here's are components on JLCPCB with this footprint: https://jlcsearch.tscircuit.com/components/list?package=SOT-343-4

There may be reference dimensions/designs on the [kicad viewer](https://tscircuit.github.io/kicad-viewer)

This issue may be eligible for a $10 bounty from the TSCircuit team (please review!)

PR #440, opened 2025-12-10
Implement SOT-343-4

shehaban:
/fixes https://github.com/tscircuit/footprinter/issues/396
- Added SOT-343-4 footprint definition with 4 pads.


seveibar (approved):
just a note, we're not accepting bounties or actively accepting anything except super-popular footprints (there are not too many left), so this is a great PR, especially for a first PR, but other repos are higher value-add for the project

techmannih (commented):
Please add it in footprinter.ts

Issue #382, opened 2025-10-14
Fix to92 screenshot (account for plated hole with square pad and pills)


techmannih:
https://github.com/tscircuit/footprinter/pull/378/files

PR #445, opened 2025-12-16
Improve TO-92 footprint pad shapes and KiCad parity

shehaban:
/fixes https://github.com/tscircuit/footprinter/issues/382
This PR improves the TO-92 footprint by refining pad geometry and improving parity with KiCad footprints.

- Use a **rectangular pad for pin 1** (via `platedHoleWithRectPad`) to clearly mark pin orientation.
- Introduce **pill-shaped pads** for remaining pins where appropriate.
- Adjust pad proportions to better match KiCad’s TO-92 footprints.
- Update snapshots and KiCad parity tests to reflect the new geometry.

### DIFF
--- a/src/helpers/boolean-difference.ts
+++ b/src/helpers/boolean-difference.ts
@@ -174,10 +174,7 @@ function elementToPolygon(element: FootprintElement): Flatten.Polygon | null {
       }
     }
 
-    if (
-      element.type === "pcb_plated_hole" &&
-      element.shape === "pill"
-    ) {
+    if (element.type === "pcb_plated_hole" && element.shape === "pill") {
       if (
         element.outer_height &&
         element.outer_width &&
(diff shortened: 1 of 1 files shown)


src/fn/to92.ts
```diff
@@ -30,13 +32,37 @@ const generateSemicircle = (
   })
 }
 
+const platedHolePill = (
```
imrishabh18: This seems redundant, also don't use `any` types

shehaban: i use it to make this shape 
image

techmannih: Yes, if it does not exist in the helper folder then you can add it there. But it's not good in this file.

techmannih:
> Need to add support platedHoleWithRectPad for boolean_differece 


 this is not resolved

techmannih:
@shehaban Sorry, I didn’t notice earlier that that was CircularHoleWithPillPad. You’ll need to add support for CircularHoleWithPillPad/platedHolePill

PR #447, opened 2025-12-18
feat: add inline support for to92l

Juste-Leo2:
Added support for the `_inline` variant to the `to92l` footprint.
- The implementation is inspired by the logic used in the `to92s` footprint.
- **Technical adjustment:** In inline mode, I reduced the pad diameter (`od`) to 1.0mm. This prevents the pads from overlapping and causing short-circuits, as the default 1.3mm diameter was too large for the 1.27mm pin pitch.
- Added a test case for `to92l_inline` to ensure correct alignment.


Juste-Leo2:
> please also add kicad parity test

Does it have a parity test? I'll take a look at that :)

Juste-Leo2:
Footprints aren't obvious xD. I have a bug in the test display for inline, and I'm stuck with simple circles for pins 2 and 3.

image

image


Do you think you're validating? Honestly, I don't think it's a big deal, and I've seen other inline skaters behave like that.

Juste-Leo2:
Here is the result between the two for the inline footprint :)

image

tests/to92l.test.ts
```diff
@@ -7,3 +7,9 @@ test("to92l", () => {
   const svgContent = convertCircuitJsonToPcbSvg(circuitJson)
   expect(svgContent).toMatchSvgSnapshot(import.meta.path, "to92l")
 })
+
+test("to92l_inline", () => {
+  const circuitJson = fp.string("to92l_inline").circuitJson()
+  const svgContent = convertCircuitJsonToPcbSvg(circuitJson)
+  expect(svgContent).toMatchSvgSnapshot(import.meta.path, "to92l_inline")
```
seveibar: there is a bug in this test snapshot, maybe due to an old version of circuit-json and/or circuit-to-svg, @techmannih can help resolve

seveibar: (the pills don't appear)

seveibar (approved):
see comment

PR #451, opened 2025-12-29
fix sot323 footprint

nailoo:


techmannih:
@nailoo Fix the silkscreen

PR #452, opened 2025-12-29
fix sot343 footprint

nailoo:


seveibar (commented):
@techmannih please make sure his kicad footprint selection is good- there may be variants

PR #454, opened 2026-01-01
Add plcc footprint

Sahil-Gupta584:
Adds PLCC footprint.

Tests is passing successfully:
image


techmannih (changes requested):
ahh you need to fix your footprint which should be exact match with kicad footprint

Screenshot_2026-01-02_07-34-43

Sahil-Gupta584:
@techmannih seems good now?

Issue #371, opened 2025-10-06
Implement PDIP-8


nailoo:
Here's are components on JLCPCB with this footprint: https://jlcsearch.tscircuit.com/components/list?package=PDIP-8

There may be reference dimensions/designs on the [kicad viewer](https://tscircuit.github.io/kicad-viewer)

This issue may be eligible for a $10 bounty from the TSCircuit team (please review!)

buildingvibes:
/attempt #371

Implemented PDIP as a footprint variant that delegates to the existing DIP implementation. PDIP (Plastic Dual In-line Package) uses the same dimensions as standard DIP: 300mil row spacing, 2.54mm pitch.

PR: #504

BossChaos:
🎯 Working on implementing this footprint! Will submit a PR with full implementation and tests. Wallet: 0xdaE5d307339074A24F579dB48e7c639359D94904

Heyzerohey:
/claim #371 via https://github.com/tscircuit/footprinter/pull/642

Implemented robust PDIP through-hole footprint (PDIP-8) in footprinter, matching the standard DIP-8 dimensions (2.54mm pitch, 300mil row spacing). All tests pass successfully.

Ojas2095:
/attempt #371

PR: https://github.com/tscircuit/footprinter/pull/650

Guciolek:
/attempt #371

Implemented in PR #655: `pdip` and `spdip` aliases for the existing `dip` footprint. Both resolve to the standard DIP layout (same 2.54mm pitch, through-hole pads, row spacing). Includes 4 test cases covering lowercase/uppercase/parameterized variants. All 416 existing tests still pass.

b3417:
Verification update: 
bun test tests/dip.test.ts tests/string-parser.test.ts passes, and un run build passes. The new PDIP/SPDIP aliases normalize to the existing DIP footprint without changing the footprint geometry.

chengdazhi:
/attempt #371

PR: https://github.com/tscircuit/footprinter/pull/702

Implements `pdip` (PDIP-8) reusing the standard DIP layout, verified against KiCad `DIP-8_W7.62mm` with a new parity test (courtyard IoU 99.90%, pads fully overlap). Full `bun test` suite passes (437/437).

JoTalbot:
[...]

PR #474, opened 2026-01-13
feat: add PDIP footprint support

Kemalyavas:
## Summary
- Add support for PDIP (Plastic Dual In-line Package) footprint string
- PDIP is a standard alias for DIP packages, commonly used in datasheets
- Supports all DIP variants: pdip8, pdip14, pdip16, pdip8_wide, etc.

## Changes
- `src/fn/pdip.ts` - New file that exports DIP as PDIP alias
- `src/fn/index.ts` - Export pdip function
- `src/footprinter.ts` - Add pdip type definition
- `tests/pdip.test.ts` - Tests for pdip8, pdip14, pdip16, pdip8_wide

## Test Plan
- [x] All 345 existing tests pass
- [x] 6 new pdip tests pass
- [x] `pdip8` parameters match `dip8` parameters
- [x] Case-insensitive: `PDIP8` resolves same as `pdip8`

Closes #371

/claim #371


rushabhcodes (changes requested):
also add kicad parity test, see others for ref

src/fn/pdip.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { dip, dip_def, extendDipDef, getCcwDipCoords } from "./dip"
+
+// PDIP uses the same implementation as DIP
+export const pdip = dip
+
+// Export the definition for type compatibility
+export const pdip_def = dip_def
+export const extendPdipDef = extendDipDef
+export const getCcwPdipCoords = getCcwDipCoords
```
techmannih: please follow code maintainability

tests/kicad-parity/pdip8_kicad_parity.test.ts
```diff
@@ -0,0 +1,18 @@
...
+      "pdip8_w7.62mm",
+      "Package_DIP.pretty/DIP-8_W7.62mm.circuit.json",
+    )
+
+  const svgContent = convertCircuitJsonToPcbSvg(combinedFootprintElements)
+  expect(svgContent).toMatchSvgSnapshot(import.meta.path, "pdip8")
+  expect(booleanDifferenceSvg).toMatchSvgSnapshot(
+    import.meta.path,
+    "pdip8_boolean_difference",
+  )
```
techmannih: Adding pdip and comparing with dip, this is incorrect

Kemalyavas:
Thanks for the feedback! You're right - I've replaced the KiCAD parity test.

PDIP (Plastic Dual In-line Package) is just an alias for DIP - they have identical physical dimensions. KiCAD library doesn't have separate PDIP footprints because they're the same as DIP.

Instead, I've added an alias verification test (`pdip7.test.ts`) that confirms `pdip8` produces identical circuit JSON as `dip8`. This is more appropriate since PDIP is purely an alias.

PR #483, opened 2026-01-18
Update SVG snapshots for TSSOP-10 footprint alignment analysis

rushabhcodes:
- Adjusted dimensions and positions of PCB pads and boundaries in tssop10.snap.svg.
- Modified silkscreen paths and text attributes for better alignment representation.
- Updated viewBox and dimensions in tssop10_boolean_difference.snap.svg for consistency.


techmannih:
@rushabhcodes why are you updating without fix.

rushabhcodes:
there was a change in kicad which cause tests to fail without any  change in footrpinter, i merged it as a temp solution as i was working on other part, will fix this soon

techmannih:
Yep, don't merge like this without fixing. If you are working with another one snapshots should be automatically updated in that pr.

PR #486, opened 2026-01-18
feat: enhance SSOP definition with additional parameters and coordinate calculations

rushabhcodes:
This pull request significantly refactors and extends the `ssop` footprint generator in `src/fn/ssop.ts`. The changes introduce a new schema for SSOP parameters, add logic for pad and silkscreen generation, and improve type safety and flexibility.

**Refactor and feature expansion for SSOP footprint generation:**

* Replaced the old `extendSoicDef`-based parameter schema with a new `zod`-based `ssop_def`, supporting more explicit and defaulted fields, improved type inference, and automatic handling of missing `pw`/`pl` values.
* Added the `getSsopCoords` function and a loop to generate rectangular pads for each pin, using the new parameter schema and helper functions for accurate pad placement.
* Introduced silkscreen generation: constructs a silkscreen border using a path and adds a silkscreen reference text, leveraging new helpers and the `u_curve` for rounded corners.
* Updated the return type of `ssop` to provide the new pad and silkscreen elements, replacing the old `soicWithoutParsing` output.
* Improved type safety by switching from `AnySoupElement` to `AnyCircuitElement` and defining a new `SsopInput` type. (F0be25


techmannih (commented):
@rushabhcodes please don’t update this until we actually need to. For the TSSOP package, you’ve used a lot of magic numbers, which is why I gave that advice.

rushabhcodes:
Screenshot_20260118_155404

PR #488, opened 2026-01-19
fix: restore pinNumber declaration in mounted PCB module for side pins handling

rushabhcodes:
This pull request makes a small but important change to the `mountedpcbmodule` function to improve how the first pin on each side is identified and handled. The change introduces a new optional parameter, `isFirstOnSide`, to the `addPin` function, and updates the logic to use this flag when adding pins in a loop.

Pin placement improvements:

* Added an optional `isFirstOnSide` parameter to the `addPin` function signature in `mountedpcbmodule.ts` to explicitly indicate when a pin is the first on its side, improving clarity and flexibility.
* Updated the pin-adding loops to pass `i === 0` as the `isFirstOnSide` argument, ensuring the first pin on each side is correctly identified and handled. [[1]](diffhunk://#diff-a9c7a9272a01d9c9a2cd519926a351e8baacedc43f26a815dfca442035dbcb29L353-R354) [[2]](diffhunk://#diff-a9c7a9272a01d9c9a2cd519926a351e8baacedc43f26a815dfca442035dbcb29L363-R364)
* Modified the condition inside `addPin` to use `isFirstOnSide` (if provided) instead of relying solely on `pinNumber === 1`, making the logic more robust for modules with multiple sides or custom pin numbering.


seveibar (changes requested):
Cant have multiple pin1

seveibar:
We should never have dup pin numbers

PR #492, opened 2026-01-31
fix the pad width and pad height for sot363

nailoo:


techmannih (changes requested):
looks like some difference 0.24%. can you fix it.

nailoo:
@techmannih I have updated and exact match with the kicad but idk why it still show 24% percent difference

PR #497, opened 2026-02-08
Fix JST PH footprint to generate correct number of pads

### LINKED ISSUE #495: jst_ph_4 footprint generates only 2 pads instead of 4
## Summary

## Root Cause

In `src/fn/jst.ts`:

   ```typescript
   ```

## Reproduction

```tsx
```

## Expected

## Suggested Fix

## Impact

BoozeLee:
- Parse pin count from footprint string (e.g., jst_ph_4 → 4 pins)
- Update generatePads() to dynamically create pads for PH variant
- Add tests for jst_ph_4 and jst_ph_6 footprints

Fixes #495

Before: jst_ph_4 only generated 2 pads (hardcoded)
After: jst_ph_4 correctly generates 4 pads
/claim #495


BoozeLee:
Hi @seveibar! 👋

  This PR fixes the bug reported in #495 where JST PH
   footprints with more than 2 pins were only
  generating 2 pads.

  The fix:
  - ✅ Parses pin count from footprint strings (e.g.,
   `jst_ph_4` → 4 pins)
  - ✅ All 356 tests passing (added 2 new tests)
  - ✅ Follows the same pattern as JST SH variant

  Would it be possible to add a bounty to #495? This
  was a clear bug with a well-defined solution, and
  I'd love to continue contributing to tscircuit! 🚀

  Thanks for maintaining such a great project!

PR #498, opened 2026-02-08
Fix JST PH footprint to support dynamic pin counts

### LINKED ISSUE #495: jst_ph_4 footprint generates only 2 pads instead of 4

aifunmobi:
## Summary
- Parse pin count from footprint string using canonical `jst5` syntax (e.g., `jst5` = JST PH with 5 pins)
- Uses the standard `num_pins` mechanism from the footprint string parser, consistent with `dip8`, `soic16`, etc.
- Dynamic pad generation loop replaces hardcoded 2-pad PH layout
- Scale silkscreen body width based on pin count and pitch

## Changes
- `src/fn/jst.ts`: PH variant now reads `num_pins` from the string parser instead of ad-hoc regex. Dynamic pad generation and silkscreen sizing for PH.
- `tests/jst.test.ts`: Tests use canonical syntax (`jst3`, `jst4`, `jst6`) with pad count assertions.

## Test plan
- [x] All 357 existing tests pass
- [x] `jst3`, `jst4`, `jst6` generate correct number of PH pads
- [x] Default `jst` still generates 2 pads (backward compatible)
- [x] SH variants (`jst_sh`, `jst_sh6`, etc.) unchanged

Fixes #495

### DIFF
--- a/src/fn/jst.ts
+++ b/src/fn/jst.ts
@@ -19,27 +19,19 @@ export const jst_def = base_def.extend({
   pl: length.optional(),
   w: length.optional(),
   h: length.optional(),
-  sh: z
-    .preprocess((v) => (v != null ? Boolean(v) : undefined), z.boolean())
-    .optional()
-    .describe(
-      'JST SH (Surface-mount) connector family. SH stands for "Super High-density".',
-    ),
-  ph: z
-    .preprocess((v) => (v != null ? Boolean(v) : undefined), z.boolean())
-    .optional()
-    .describe(
-      'JST PH (Through-hole) connector family. PH stands for "Pin Header".',
-    ),
+  sh: z.coerce.boolean().optional().describe("JST SH (Surface-mount) connector family."),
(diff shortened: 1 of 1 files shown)


seveibar:
Whoever did the other syntax may have messed up jst${number} is the right syntax

aifunmobi:
Good catch! Updated to support the `jst4_ph` syntax as the preferred format. Now supports all three styles:

- `jst4_ph` — preferred (number after component name)
- `jst4` — defaults to PH variant
- `jst_ph_4` / `jst_ph4` — also work as fallbacks

Added tests for both `jst4_ph` and `jst4`. All 359 tests pass.

rushabhcodes (commented):
You can also join our [discord](https://discord.gg/W57ByBus)

PR #584, opened 2026-04-07
Fix: courtyards for quad to use body based approach

MustafaMulla29:


seveibar (commented):
needs some cleanup, but it's better

seveibar (changes requested):

seveibar (approved):

PR #588, opened 2026-04-07
feat(vssop): switch to generic body-based courtyard

MustafaMulla29:


seveibar (approved):
see comment, i would advise making a function createUnionOfRects which outputs a polygon

PR #590, opened 2026-04-08
Fix courtyards for jst, sot, to220, vson, electrolytic

MustafaMulla29:


src/fn/jst.ts
```diff
@@ -49,6 +49,35 @@ export type jstDef = z.input<typeof jst_def>
...
+  width: number,
+  height: number,
+) => {
+  const halfX = width / 2
+  const halfY = height / 2
+  bounds.minX = Math.min(bounds.minX, centerX - halfX)
+  bounds.maxX = Math.max(bounds.maxX, centerX + halfX)
+  bounds.minY = Math.min(bounds.minY, centerY - halfY)
+  bounds.maxY = Math.max(bounds.maxY, centerY + halfY)
+}
```
seveibar, line 85: Two parameter rule, also poorly named.

modifyBoundsToIncludeRect

always make sure effects are clear in naming

src/fn/vson.ts
```diff
@@ -69,20 +70,52 @@ export const vson = (
...
-  const courtyardPadding = 0.25
-  const centerY = ((num_pins / 2 - 1) * p) / 2
-  const crtMinX = -(w / 2 + pinw / 2 + courtyardPadding)
-  const crtMaxX = w / 2 + pinw / 2 + courtyardPadding
-  const crtMinY = -(Math.max(grid.y / 2, centerY + pinh / 2) + courtyardPadding)
-  const crtMaxY = Math.max(grid.y / 2, centerY + pinh / 2) + courtyardPadding
-  const courtyard: PcbCourtyardRect = {
-    type: "pcb_courtyard_rect",
-    pcb_courtyard_rect_id: "",
+  const roundToCourtyardGrid = (value: number) =>
```
seveibar: This function doesnt make sense. What is a courtyard grid? I should have said something earlier but there is no suvh thing as a courtyard grid. You could say roundTo10um but why are we rounding anyway?

seveibar (approved):
See comments.

We never use inline closures, i keep seeing them in your code! They are an antipattern and prevent refactoring code because they steal scope!

PR #591, opened 2026-04-08
Update remaining courtyard geometry and add parity tests

MustafaMulla29:


seveibar (changes requested):
Create a method for creating a capsule outline

src/fn/pinrow.ts
```diff
@@ -351,17 +351,29 @@ export const pinrow = (
   // Add centered silkscreen reference text
   const refText: SilkscreenRef = silkscreenRef(0, p, 0.5)
 
-  const halfSpanX = ((numPinsPerRow - 1) / 2) * p
   const padHalfX = parameters.smd ? parameters.pw / 2 : od / 2
   const padHalfY = parameters.smd ? parameters.pl / 2 : od / 2
-  const courtyardPadding = 0.25
+  const roundToCourtyardGrid = (value: number) =>
```
seveibar: Remove,ALL instances of this inline closures ahould not exist, this antipattern is now everywhere. There is no such think as a courtyard gri

seveibar (changes requested):
Ehat is a halfX? Do you mean halfWidth? What is halfY?

seveibar (approved):

PR #597, opened 2026-04-17
feat: add PLCC, SOJ, led2835, led5050, and lpcc footprints

Angelebeats:
This PR adds PLCC, SOJ, led2835, led5050, and lpcc footprints to the footprinter. All footprints are implemented using current standards and verified with local tests.


Angelebeats:
??????????,??????????????????,??!

PR #611, opened 2026-04-28
fix: parse pin count from sh variant in JST footprints

64johnlee:
## Summary

Fix Issue #495 - jst_ph_4 footprint generates only 2 pads instead of 4

**Problem:** The JST footprint parser only parsed pin count from the string for ZH variant, not for SH variant. This meant `jst_ph_4` would fail to parse the 4 pins.

**Solution:** Added `shMatch` regex to parse pin count from SH variant string, e.g., `jst_ph_4` extracts 4 pins.

```typescript
const shMatch = str.match(/(?:^|_)sh(\d+)(?:_|$)/)
if (shMatch && shMatch[1]) {
  numPins = Number.parseInt(shMatch[1], 10)
}
```

## Testing

- `jst_ph_4` now correctly generates 4 pads
- `jst_ph_2` still works correctly (2 pads)

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,12 +1,10 @@
 {
-  "$schema": "https://biomejs.dev/schemas/1.7.3/schema.json",
-  "organizeImports": {
-    "enabled": true
-  },
+  "$schema": "https://biomejs.dev/schemas/2.4.13/schema.json",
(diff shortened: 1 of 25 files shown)


techmannih2 (approved):
fix format check, rest lgtm

PR #612, opened 2026-04-30
fix: align passive silkscreen geometry with KiCad standards

techmannih:
Silkscreen fixes mainly for passive footprints, but now silkscreen comparison is enabled and verified for all footprints in the parity test suite.

Precision Geometry: Replaced heuristic silkscreen sizing with exact KiCad-derived values for all standard passive sizes (0402 through 2512).

Visual Parity: Refactored silkscreen paths to use the KiCad-standard dual horizontal line style.


src/helpers/passive-fn.ts
```diff
@@ -21,6 +21,8 @@ type StandardSize = {
   w_mm_min: number // body width
   courtyard_width_mm?: number
   courtyard_height_mm?: number
+  ss_w?: number // silkscreen length
+  ss_y?: number // silkscreen y center
```
seveibar: properties aren't allowed to have `_`

seveibar: oh this some dumb variable naming but i guess you're matching the existing bad pattern- these aren't properties but are styled as properties

seveibar (changes requested):
really bad change, all capacitors lose polarity, we're also not interested in copying kicad completely

PR #613, opened 2026-04-30
fix: update 0402 resistor silkscreen to use dual-line geometry

techmannih:


seveibar (approved):

seveibar:
"polarized vs nonpolarized" is how you should refer to these silkscreen differences, it's not clear from the code

PR #656, opened 2026-06-13
Fix diode silkscreen polarity marking to match cathode orientation

Abse2001:


seveibar (changes requested):
diodes are not a passive component

seveibar:
passive should not be used for diodes

seveibar (commented):
i'm super confused for a couple reasons

seveibar:
@Abse2001 so we're just flipping pin polarity for LEDs and Diodes??? why???? This is a big potential footgun if we're not super clear about this

seveibar:
what does the IPC standard say. The fix is more likely a core mapping issue. Find the IPC spec. See what it says pin1 should be. Find what it says the silkscreen should be. Everything you do should try to be in compliance with the IPC spec

seveibar:
apparently some people call diodes passives so we can possible use it but not if we're changing standard pin definitions

seveibar:
image

seveibar:
I would also set up kicad comparison screenshots with silkscreen to keep us conventional. We should have a "+" indicator to make sure that there isn't confusion

seveibar:
@Abse2001 sorry for the initial reaction- so IPC spec seems to be to make pin1 = cathode (which is negative), we just need to be very careful because these changes end up breaking lots of circuits- i'm very happy we're getting to the bottom of this but we should use kicad snapshots to make sure we have a way to prevent these bugs and prove correctness in the future

Abse2001:
@seveibar this mean we need to change the port mapping for diode to be cathode pin1, is it ok to keep it as passive?

seveibar:
@Abse2001 i think we should create a new fn for this type of component and change the silkscreen- the passive fn isnt explicit (since diodes are not always considered passive) and we could have more consistent handling. We can always import dimensions for 0402 etc but yea i dont like how you use fn (which we should avoid using- it is an anti-pattern where you pass in a parameter that changes a function, called control coupling or mode parameters, see martin fowler)

Tldr, we should not dramatically alter behavior implicitly via a parameter to passive

Abse2001:
@seveibar can I get another review, and the port flip will be in core right ?

PR #657, opened 2026-06-14
fix: scale JST PH silkscreen body dynamically with pin count (#495)

### LINKED ISSUE #495: jst_ph_4 footprint generates only 2 pads instead of 4

sagarmaurya64-ai:
This PR resolves #495 where the silkscreen body box for the JST PH (through-hole) variant remained hardcoded to a 2-pin size (6mm x 5mm) even when the footprint has 4 or more pins.

### Implementation Details:
1. **Dynamic Silkscreen Outline**: Modified `generateSilkscreenBody` in `src/fn/jst.ts` to calculate the JST PH silkscreen width dynamically as `(numPins - 1) * p + 3.8mm` (where `1.9mm` clearance is applied on each side). This matches the physical connector width and maintains 100% backward compatibility with the existing 2-pin JST PH snapshot (`jst2_ph`).
2. **Added Unit Tests**: Added a unit test verifying that the silkscreen bounding box width scales correctly for different pin counts (`jst_ph_4` vs `jst2_ph`).
3. **Updated Snapshot**: Re-generated the snapshot for `jst_ph_4` to match the newly scaled, correct dimensions.

Closes #495


sagarmaurya64-ai:
Hi @seveibar! 👋 Conflict resolved and tests are passing, please review and merge.

sagarmaurya64-ai:
Hi @seveibar! 👋 Just checking in to see if you have any feedback or if there is anything else needed to merge this. The conflicts are resolved and all CI checks are green. Thank you! 🙏

sagarmaurya64-ai:
Hi @seveibar, friendly follow-up! This PR is ready for review - all tests pass locally. Would appreciate your feedback when you get a chance. Happy to address any changes needed. Thank you!

PR #659, opened 2026-06-15
Enable silkscreen on parity comparisons in KiCad footprint validation fixtures

Abse2001:


imrishabh18 (commented):
Why is diff before and after still 0?

Abse2001:
@imrishabh18 I believe we want the silkscreen diff to visual only, so we can compare to kicad, not sure if we want to change our silkscreen to match there's 
CC @seveibar

imrishabh18 (approved):

PR #668, opened 2026-06-29
fix: make the sot343 footprint symmetric

### LINKED ISSUE #667: sot343 footprint is lopsided — the right pads sit too close to the center
I was looking through the SOT footprints and `sot343` seems to be the odd one out — its pads aren't symmetric.

If you generate it and check the pad X positions:

```
sot343:  [-1.056, -1.056, 0.55, 0.55]   center = -0.253   columns ~1.6mm apart
sot323:  [-0.891, -0.891, 0.891]        center =  0        (symmetric)
sot363:  [-0.84, -0.84, -0.84, 0.84, …] center =  0        (symmetric)
```

The left column is at -1.056 but the right column only reaches +0.55, so the whole part is shifted off its origin and the two rows of pads end up ~1.6mm apart instead of ~2.1mm. Every other SOT footprint is symmetric about the center, so this one looks wrong — the land pattern wouldn't match a real SC-70-4 part, and the right-hand leads would miss their pads.

It looks like a small typo in `getCcwSot343Coords` (`src/fn/sot343.ts`): pins 1 and 2 use `x: -p * 1.92`, but pins 3 and 4 just use `x: p` — the `* 1.92` is missing on the right side. The hardcoded courtyard outline just above it is shifted to match those off-center pads too.

I'll open a PR.

abdalraof-albarbar:
`sot343` was the only SOT footprint with lopsided pads. Pins 1 and 2 are placed at `x = -p * 1.92`, but pins 3 and 4 were placed at just `x = p`, so the right column sat much closer to the center than the left:

```
before:  pads x = [-1.056, -1.056, 0.55, 0.55]    (center -0.253, columns ~1.6mm apart)
after:   pads x = [-1.056, -1.056, 1.056, 1.056]  (centered, columns ~2.1mm apart)
```

So the part was off-center and the spacing didn't match a real SC-70-4. I gave the right column the same `* 1.92` spacing as the left so the four pads mirror each other.

One catch: the courtyard outline was hardcoded to the old off-center pads (it ran from `-1.703` to `+1.197`), so once the right pad moved outward it would have stuck outside the courtyard. I mirrored the right side of the outline too, so it stays symmetric and still wraps the pads. The silkscreen already centers itself off the pad bounds, so that fixed itself.

Regenerated the two sot343 snapshots.

Closes #667


abdalraof-albarbar:
Heads-up on the red `test` check — it isn't from this PR. All the `parity/*` tests fail the same way (`sot89`, `bga165`, `soic24` … none of which are touched here), because the cached KiCad reference now stores each courtyard as a set of 2-point line segments, while the parity fixture only builds a polygon from an outline with 3+ points. So `kicadPolygon` ends up empty and it throws "Could not convert courtyard geometry into polygons". `main` was green a few days ago, before the cache changed format.

The sot343 tests themselves pass. Once the parity comparison can read the reference again I'll regenerate the two kicad-parity sot343 snapshots — the fix should line up with the KiCad SOT-343 reference more closely (its courtyard is symmetric at ±1.45, which is exactly what this change is about).

abdalraof-albarbar:
Reworked this against the now-restored parity test. The original version made the pads symmetric but at the wrong position (`p*1.92` = ±1.056). The KiCad SC-70-4 reference actually has its pads at **±0.825** (`p*1.5`) and a **±1.45** courtyard, so I matched those — the parity courtyard diff is now **0%**. Rebased onto current `main` (which has the courtyard-segment stitcher) and regenerated the sot343 snapshots, including the kicad-parity ones.

PR #670, opened 2026-06-29
test: repro for sot89 not matching the KiCad reference (repro only, no fix)

abdalraof-albarbar:
**Repro only — this PR does not include the fix.** It documents the bug with a `test.failing` so it's recorded and CI stays green. A follow-up PR will fix the footprint.

### The bug
`sot89` doesn't match the real SOT-89-3 land pattern. Two problems:
- The **collector tab (pin 2)** should be the large polygon land the KiCad reference uses (~3.1 × 1.7mm); footprinter renders it as a small rect (1.475 × 0.9mm).
- The **lead pads** sit at `x = -2.19` vs KiCad's `-1.95`.

Together the footprint is **~47% off the KiCad courtyard** — a SOT-89 part soldered to it would have an inadequate collector/thermal tab.

### Why it wasn't caught
The existing `parity/sot89` test only does `toMatchSvgSnapshot`, so the mismatch is baked into the snapshot and never asserted. (sot343's parity test, by contrast, asserts `courtyardDiffPercent < 5`.) This adds that assertion as `test.failing`:

```ts
test.failing("parity/sot89 courtyard matches KiCad within 5%", async () => {
  const { courtyardDiffPercent } = await compareFootprinterVsKicad(
    "sot89", "Package_TO_SOT_SMD.pretty/SOT-89-3.circuit.json",
  )
  expect(courtyardDiffPercent).toBeLessThan(5)  // currently 46.9%
})
```

### Follow-up
A separate fix PR will make pin 2 a large polygon tab and move the lead pads to ±1.95 to match the reference, then flip this to a normal `test()`.

### DIFF
--- a/tests/kicad-parity/sot89_kicad_parity.test.ts
+++ b/tests/kicad-parity/sot89_kicad_parity.test.ts
@@ -2,12 +2,20 @@ import { expect, test } from "bun:test"
 import { compareFootprinterVsKicad } from "../fixtures/compareFootprinterVsKicad"
 import { convertCircuitJsonToPcbSvg } from "circuit-to-svg"
 
-test("parity/sot89", async () => {
-  const { combinedFootprintElements, booleanDifferenceSvg } =
-    await compareFootprinterVsKicad(
-      "sot89",
-      "Package_TO_SOT_SMD.pretty/SOT-89-3.circuit.json",
-    )
+// sot89 doesn't match the real SOT-89-3 land pattern: the collector tab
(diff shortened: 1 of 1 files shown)


seveibar (commented):
modify the other test to have the assertion and put test.failing on the original test, we try to keep one test per file

abdalraof-albarbar:
Good call — folded the `courtyardDiffPercent` assertion into the existing `parity/sot89` test and marked it `test.failing`, so it's one test per file now. @seveibar

seveibar (approved):

PR #671, opened 2026-06-30
fix: match the KiCad SOT-89-3 land pattern (large collector tab)

abdalraof-albarbar:
Follows up #670 (the repro).

`sot89` placed its lead pads at `x=-2.19` (vs KiCad's `-1.95`) and rendered the collector tab (pin 2) as a tiny 1.475×0.9 rect instead of the large tab the real SOT-89-3 land pattern uses — leaving the footprint **~47% off** the KiCad courtyard. A SOT-89 part soldered to it would have an inadequate collector/thermal tab.

This fix:
- puts the two lead pads at **±1.95**,
- makes pin 2 the **large collector tab** (~3.1 × 1.7mm, centred at x≈0.435),
- squares the courtyard to the KiCad reference.

Courtyard diff is now **0%**. Regenerated the snapshots and flipped `parity/sot89` from `test.failing` to a passing `test()`.


seveibar (changes requested):
this does not seem correct, look at a direct kicad reference (open sot89 in kicad) and screenshot, the data you're comparing to looks invalid

abdalraof-albarbar:
Sure, I’ll check it in KiCad.

abdalraof-albarbar:
Checked it directly in KiCad (rendered SOT-89-3 with `kicad-cli`) — you're right, the data I was comparing to is invalid.

Pin 2 (the collector tab) is an `smd custom` pad: an anchor rect (`1.475×0.9`) plus a polygon. KiCad draws the union, which is a T-shape — a narrow neck reaching toward the leads, then the wide tab. But `kicad-to-circuit-json` drops the anchor and keeps only the polygon, so the cached reference (and my footprint) lost the neck.

So it's a converter bug, not just a footprint one — repro up at tscircuit/kicad-to-circuit-json#157, with a fix ready (emit the anchor shape, skip KiCad's near-zero placeholder anchors). Once that lands I'll update this PR's sot89 to match the real KiCad tab (with the neck). @seveibar

abdalraof-albarbar:
Update: the converter fix merged (tscircuit/kicad-to-circuit-json#158), but kicad-mod-cache still serves the old SOT-89-3 conversion (tab without the neck, pad 2 stops at x≈-1.13). Once the cache regenerates I'll update the sot89 geometry here to the true T-shaped tab and re-run parity.

abdalraof-albarbar:
Superseded by #689 — that landed the SOT-89-3 collector as a proper polygon land pattern with the neck reaching x=-2.6, which matches the (now regenerated) KiCad reference. My rect-based version here is redundant and would actually regress it, so closing. Thanks!

PR #672, opened 2026-07-02
Add pinLabelSide option to pinrow

derekbreden:
## What

Adds `pinLabelSide` to `pinrow` — specify which side of the pin row the pin labels sit on: `"top" | "bottom" | "left" | "right"`.

In the string DSL it's set via flags mirroring `mountedpcbmodule`'s `pinRowSide` / `pinrow<side>` convention: `pinlabeltop`, `pinlabelbottom`, `pinlabelleft`, `pinlabelright` (e.g. `pinrow5_pinlabelbottom`), which resolve to the canonical `pinLabelSide` enum.

## Behavior

`pinLabelSide` overrides the anchor side otherwise derived from text alignment / `pinlabelorthogonal`; unset → output unchanged. It's distinct from the layer option `bottomsidepinlabel`, and is scoped to the pin labels (ref-des unchanged).

- `pinrow5` (default) — labels above the row.
- `pinrow5_pinlabelbottom` — labels below the row.

## Tests

`pinrow5_pinlabelbottom` asserts the `pinlabelbottom` flag resolves to `pinLabelSide: "bottom"` and that a pin label lands on the opposite side vs the default, plus an SVG snapshot. `bun test` / `build` / `biome` all green.

## Note

Supersedes the earlier `flippinlabels` approach per review — naming the absolute side rather than an axis-ambiguous inversion.


seveibar:
Generally we try not to specify things with inversions, flip is sort of axis-ambiguous. Can you think of a way to specify this position where the behavior is more obvious?

Eg flip often refers to top/bottom LAYER

derekbreden:
Good call — agreed on both counts (the inversion, and "flip" reading as a top/bottom *layer* op).

Dropped `flippinlabels` and replaced it with an explicit **`pinLabelSide`** (`"top" | "bottom" | "left" | "right"`), following the `mountedpcbmodule` `pinRowSide` / `pinrow<side>` convention already in the repo: string-form flags `pinlabeltop` / `pinlabelbottom` / `pinlabelleft` / `pinlabelright` resolve to the canonical enum (e.g. `pinrow5_pinlabelbottom`). It names the position directly (no inversion), and sits alongside the layer option `bottomsidepinlabel` the same way `pinrowbottom` does in mountedpcbmodule. Overrides the alignment-derived side; unset leaves output unchanged. PR + test/snapshot updated.

PR #674, opened 2026-07-02
Fix TO-92 footprint origin

Lathikaa-S:
image

## Issue
The TO-92 footprint was not centered around `pcbX: 0` / `pcbY: 0`. The body and courtyard sat above the origin, so the component/model origin did not line up cleanly with the footprint. This made the footprint look offset when compared with the 3D/model placement.

## What changed
- Moved the TO-92 body so its silkscreen is centered on `0,0`.
- Moved the courtyard to stay centered around the same origin.
- Placed the pads relative to the centered body, instead of applying a broad post-processing offset.
- Kept the implementation close to the existing footprinter pattern with simple coordinate changes.
- Added a regression test that checks TO-92 body and courtyard centers are at the footprint origin.
- Updated TO-92 SVG snapshots.


mohan-bee:
image


image


i feel like the footprint moved in opposite way !!

tests/__snapshots__/to92.snap.svg
imrishabh18, line 1: I am confused what is happening here, make things clear for the reviewer. The 3d snapshot should be using this footprint and not the inline one

PR #680, opened 2026-07-03
use square pin1 pad for electrolytic capacitor

rushabhcodes:
## Summary

- use a square plated pad for electrolytic pin 1
- retain the circular plated pad for pin 2
- update the footprint and KiCad parity SVG snapshots

## Why

The generated electrolytic footprint used circular pads for both pins, while the matching KiCad footprint uses a square pad for pin 1 as its polarity marker. This caused an 8.35% pad-area difference in the parity visualization.

## Impact

Electrolytic footprints now match the KiCad pin-1 pad geometry without changing pitch, drill size, courtyard, or pin-2 geometry.

## Validation

- `bun test tests/electrolytic.test.ts tests/kicad-parity/electrolytic_kicad_parity.test.ts`
- `bun run build`


seveibar (approved):
see my title change, kicad parity isn't a goal, just a way to check results

PR #683, opened 2026-07-03
Fix pushbutton KiCad hole parity

rushabhcodes:
## Summary

- use `id` for the pushbutton drill diameter
- use `od` for the outer pad diameter
- update the KiCad parity fixture parameters and snapshots

## Why

The pushbutton generator ignored `od` and derived the outer diameter as `id × 1.5`. The parity fixture therefore inflated the drill diameter to obtain the correct 2 mm outer pad, leaving every drill larger than the matching KiCad footprint.

## Impact

Pushbutton footprints now honor both diameter parameters independently, and the parity fixture matches KiCad’s 1.1 mm drill with a 2 mm outer pad.

## Validation

- `bun test tests/pushbutton.test.ts tests/kicad-parity/pushbutton_kicad_parity.test.ts`
- `bun run build`


seveibar (commented):
the plated holes are worse now imo- there is almost no plating- i don't agree with kicad's version

PR #689, opened 2026-07-03
Fix SOT-89-3 KiCad parity

rushabhcodes:
## Summary

- match the SOT-89-3 lead pads, collector polygon, and courtyard to KiCad
- move the reference designator above the footprint instead of over the collector pad
- regenerate the affected SVG snapshots and restore the parity test as a normal passing test

## Root cause

The SOT-89-3 collector was modeled as a small rectangular pad and the outer pads and courtyard were offset from the KiCad land pattern. This produced a 46.9% courtyard difference and a misleading parity snapshot.

## Impact

The generated SOT-89-3 footprint now overlays the KiCad reference with 100% courtyard IoU, while keeping the reference text clear of copper pads.

## Validation

- `bun test tests/sot89.test.ts tests/kicad-parity/sot89_kicad_parity.test.ts`
- `bun run build`
- `bunx biome check src/fn/sot89.ts`


seveibar (commented):
Can you double check in kicad manually? I think something is off-

seveibar:
I think we’re not converting the footprint properly- it should be a builtin- easy to quickly pull up

src/fn/sot89.ts
```diff
@@ -38,6 +38,18 @@ const sot89_5CenterPadPoints = [
   { x: 0.4, y: 1 },
 ]
 
+// KiCad pad 2 is the union of its anchor rectangle and polygon primitive.
```
seveibar: Dont include kicad references in footprinter, oart goal is IPC not kicad and this misleads

seveibar (approved):

PR #691, opened 2026-07-04
feat: add rounded pad corners to all footprints

anil08607:


seveibar (changes requested):
can't merge until core is more up to date with footprinter, you can't just change thousands of footprints for the thousands of tscircuit users without shipping some of the changes to users to verify it doesn't break things-

PR #692, opened 2026-07-04
Fix KiCad boolean-difference rendering for rounded and circular SMT pads

rushabhcodes:
## Summary

Fixes incorrect SMT pad geometry in KiCad boolean-difference visualizations.

## Problem

The boolean-difference converter treated SMT pads with width and height as plain rectangles. This caused:

- KiCad `roundrect` pads to lose their corner radius
- Circular BGA pads to render as squares
- Shape inaccuracies to be hidden in parity comparisons

## Changes

- Build exact rounded-rectangle polygons using line segments and quarter-circle arcs
- Preserve and clamp `corner_radius` values
- Handle circular and polygon SMT pads before the rectangular fallback
- Regenerate affected KiCad parity snapshots

## Impact

Boolean-difference snapshots now accurately represent real KiCad pad geometry, making footprint parity regressions easier to detect and trust.

## Verification

- 108 KiCad parity tests pass
- Production build and declaration generation pass
- Affected snapshots cover rounded resistors, diodes, QFP/QFN packages, and circular BGA pads


rushabhcodes:
I’ll fix the remaining regression in a follow-up PR.

techmannih (commented):
@rushabhcodes I think @anil08607 is working on that one, he send a message in the contributor channel.

techmannih:
@rushabhcodes It would be good for us if a new contributor wants to work on it. We should encourage and help them instead of blocking them. That will motivate more people to contribute and benefit the project in the long run..

seveibar (approved):

PR #695, opened 2026-07-04
feat: add pdip aliases for dip footprints

b3417:
Adds PDIP/SPDIP aliases that normalize to the existing DIP footprint implementation, with tests for PDIP-8 parsing and parser stability.\n\n/claim #371


b3417:
Rebased onto the latest `main` — the conflict in `src/footprinter.ts` is resolved (the new `d2pak`/`to-252`/`to-263` aliases and the `pdip` aliases coexist, none were dropped). Full suite passes locally: 506/506. It's ready for review whenever you have time. If you'd prefer a different approach, or if this is no longer relevant, just say so and I'll close it myself.

PR #697, opened 2026-07-06
fix: correct pin numbering for sot723, msop, son, and make sot343 symmetric (#686, #677, #667)

### LINKED ISSUE #667: sot343 footprint is lopsided — the right pads sit too close to the center

### LINKED ISSUE #677: msop and son footprints number the right-column pins in reverse (pins 5↔8, 6↔7 swapped on 8-pin parts)
### What happens

`msop` and `son` place their right-column pins top→bottom, the same direction as the left column. On `msop8`/`son8` that puts pin 5 at the **top**-right and pin 8 at the **bottom**-right.

```ts
fp.string("msop8").circuitJson()
// pin 5 → (2, +0.975)   top-right ❌
// pin 8 → (2, -0.975)   bottom-right ❌
```

### What should happen

Dual-row packages are numbered counter-clockwise: pin 1 top-left, down the left side, then **up** the right side, so the last pin sits across from pin 1. That's what every other dual-row footprint in this repo generates:

```ts
fp.string("soic8").circuitJson()
// pin 5 → (2.15, -1.905)   bottom-right ✅
// pin 8 → (2.15, +1.905)   top-right ✅
```

`soic`, `dfn` and `sop8` use `getCcwSoicCoords` (the right column starts at `-h/2` and goes up), and `tssop`/`ssop` do the same in their own coord functions. Only `msop` and `son` differ.

### Effect

The pad positions look fine in the SVG previews (no pin numbers rendered there), but the pin→pad mapping is mirrored, so a real chip routed with these footprints gets pins 5↔8 and 6↔7 (msop8/son8), 4↔6 (son6), and 6↔10, 7↔9 (msop10) connected to the wrong nets. Affects all variants: `msop8/10/12/16`, `son6/8` (incl. `_ep`).

### Cause

`getMsopCoords` in `src/fn/msop.ts` and `getSonPadCoord` in `src/fn/son.ts` (identical copies) compute

```ts
const rowIndex = (pn - 1) % half
const row = (half - 1) / 2 - rowIndex // same for both columns
```

so both columns run top→bottom. (`vssop.ts` has the same helper but compensates by remapping pin numbers before calling it, which is why `vssop` comes out correct.)

Happy to submit a PR for this.

### LINKED ISSUE #686: sot723 footprint numbers pins 1 and 3 on the wrong pads (1↔3 swapped vs datasheet)
### What happens

The `sot723` footprint places its pads at the correct positions, but assigns the pin numbers 1 and 3 to the wrong pads:

```ts
fp.string("sot723").circuitJson()
```

| Pin | Actual (current main) | Expected (KiCad / Toshiba datasheet) |
|-----|----------------------|--------------------------------------|
| 1 | `(+0.575, 0)` — alone on the right | `(-0.575, +0.4)` — top-left |
| 2 | `(-0.575, -0.4)` — bottom-left ✅ | `(-0.575, -0.4)` — bottom-left ✅ |
| 3 | `(-0.575, +0.4)` — top-left | `(+0.575, 0)` — alone on the right |

So pins 1 and 3 are swapped. A real SOT-723 part placed on this footprint gets cross-wired (e.g. for a transistor, base and collector swap).

### Expected

The standard 3-lead SOT arrangement, which every other 3-pin SOT in this repo already follows (`sot23`, `sot23w`, `sot323`): **pin 1 top-left, pin 2 bottom-left, pin 3 alone on the right**.

This is also exactly what KiCad's official `SOT-723.kicad_mod` does (pad 1 at `(-0.575, -0.4)` in KiCad's y-down coords = top-left, pad 3 at `(0.575, 0)`). The KiCad footprint references the Toshiba datasheet, and the pad size (`0.45 × 0.4`) and pitch (`±0.575`) in `src/fn/sot723.ts` match it exactly — only the numbering differs.

### Why the KiCad parity test doesn't catch it

`tests/kicad-parity/sot723_kicad_parity.test.ts` compares the copper shapes via boolean difference, which is blind to pin numbers — swapped numbering produces identical copper.

(Found while comparing pin-numbering conventions across sibling footprints, same class of bug as #677.)

sagarmaurya64-ai:
This PR resolves #686, #677, and #667 by correcting the pin coordinate mapping logic for `sot723`, `msop`, and `son` footprints to adhere to standard layouts, and fixing the lopsided `sot343` footprint.

### Changes Details:
1. **SOT-723 Footprint Pin Mapping** (fixes #686):
   - Corrected `getCcwSot723Coords` in `src/fn/sot723.ts` to map coordinates correctly: Pin 1 (top-left), Pin 2 (bottom-left), Pin 3 (alone on the right). This resolves the previous swap of Pins 1 and 3.

2. **MSOP and SON Footprints Pin Mapping** (fixes #677):
   - Corrected `getMsopCoords` in `src/fn/msop.ts` and `getSonPadCoord` in `src/fn/son.ts` so that the right column pins are numbered counter-clockwise (counting upwards rather than downwards). This resolves the swapped/reversed pins on the right side (e.g. 5↔8, 6↔7 on 8-pin variants).

3. **SOT-343 Footprint Symmetry** (fixes #667):
   - Corrected `getCcwSot343Coords` in `src/fn/sot343.ts` to multiply the right column coordinates by `1.92` (matching the left side) so that the pads are symmetric about the center.
   - Updated `sot343CourtyardOutline` to make the courtyard outline symmetric as well.

### Verification:
- All local unit tests passed successfully (`bun test`).
- Re-generated SVG snapshots for `sot343` and `sot343_pl1.2_pw0.9_p2_w5.2_h5` to match the correct, symmetric geometries.

Closes #686
Closes #677
Closes #667


sagarmaurya64-ai:
Hi @seveibar, friendly follow-up! This PR is ready for review - all tests pass locally. Would appreciate your feedback when you get a chance. Happy to address any changes needed. Thank you!

PR #701, opened 2026-07-11
feat: add rounded pad corners to all footprints

anil08607:


imrishabh18 (commented):
I don't know if we need these kind of changes right now

anil08607:
> I don't know if we need these kind of changes right now

hey @imrishabh18, I added corner support in the boolean difference snapshot, which changed the pads. now I'm updating all the rounded corner pads to match the boolean difference snapshot.

techmannih (approved):
Resolve merge conflicts

PR #712, opened 2026-07-19
Support configurable USB-C receptacle footprints with custom pads and mounting holes

Abse2001:
image

After 
image


seveibar (changes requested):
Did you see the existing usbc variant? Can you narrow by using specific variants for fns?

Eg smdusbc

There are too many params, and some params are FAR too complex!!!!

seveibar:
Do not introduce speculative params

seveibar (approved):

PR #718, opened 2026-07-22
repro(dip): add fabrication note pin label rendering repro

itisrohit:
DIP-8 fabrication note pin label rendering bug.

The snapshot shows unresolved `{pin1}`…`{pin8}` placeholders in fabrication notes.


techmannih (commented):
Don't add repro. You can fix it directly if the snapshot exist in the repo

seveibar (changes requested):
The repro already seems to exist

itisrohit:
Closing this since the bug is fixed directly in #719.

PR #719, opened 2026-07-22
fix(dip): render fabrication note pin numbers directly

itisrohit:
Render numeric pin labels directly in DIP fabrication notes instead of unresolved `{pinN}` placeholders.

Updates the existing DIP snapshots to reflect the corrected output.


techmannih (commented):
That's fix is not good, you have to change the position of pins instead of converting them in number

itisrohit:
@techmannih 
Thanks for the review!

Just to make sure I understand correctly: are you referring to the placement/alignment of the fabrication note pin labels, rather than the `{pinN}` text itself?

I originally changed `{pinN}` to numeric labels because I couldn't find any code path that resolves `{pinN}` placeholders before rendering. If the intended fix is to adjust the label positioning while keeping `{pinN}`, I'm happy to make that change instead.

Could you clarify which positioning issue you had in mind?

techmannih:
Yes you have to change the position for {pinN} instead converting them in numbers pin label

itisrohit:
@techmannih 
Thanks! Just to make sure I implement the intended fix correctly:

Do you mean the position of the `{pinN}` labels relative to the pads (e.g. alignment/offset), or that they should be placed somewhere else entirely?

I want to preserve the existing `{pinN}` behavior and only adjust the positioning as intended.

itisrohit:
Closing this since the issue is being addressed in #720.

PR #724, opened 2026-07-24
fix(pinrow): prevent pin labels from overlapping plated holes

anil08607:
PR https://github.com/tscircuit/sparkfun-boards/pull/311

Fixes pin-row silkscreen labels overlapping plated holes, including orthogonal and vertically inverted labels.


techmannih (commented):
Can you add the more examples where the pinrows should be vertical in column and the text should be in  left and right side

techmannih (approved):

seveibar (approved):

PR #728, opened 2026-07-24
Add parameterized FPC footprints

seveibar:
## What changed

- Add a parameterized `fpcN` footprint family for FFC/FPC connectors.
- Support a single contact row with two mechanical mounting pads.
- Support alternating staggered contact rows with configurable row pitch and per-row pad lengths.
- Expose contact pitch/size and mounting-pad position/size parameters through both string and builder APIs.

Example:

```text
fpc12_p0.5mm_pw0.3mm_pl1.25mm_mpx8.88mm_mpy2.575mm_mpw2mm_mpl2.5mm
```

## Why

The JLC5000 95% IoU benchmark currently misclassifies these connector patterns as unrelated `tssop`, `dfn`, or `solderjumper` footprints. A dedicated family represents a missing topology and can cover multiple pin counts and pitches rather than special-casing one component.

The tests reproduce representative copper geometry from:

- JLCPCB C2856799, FPC-05F-12PH20
- JLCPCB C262505, AFC11-S30ICA-00
- JLCPCB C5343246, FPC-0.3HF-31PWBH10

This gives `circuit-json-to-footprinter` a reusable family to discover in a follow-up change.

## Validation

- `bun test` — 468 passed, 0 failed
- `bun run build`
- `git diff --check`
- Visual SVG snapshots added and inspected for all three representative layouts


seveibar:
CC @Abse2001

PR #752, opened 2026-07-29
repro: solder jumper bridge trace endpoint lands outside pad boundary (FP precision)

itisrohit:
`src/fn/solderjumper.ts:56-57` computes bridge trace endpoints at `padWidth / 2` from pad center (exactly on the pad edge). With non-exact IEEE 754 values like `pw=0.6604`, the compound FP error pushes the endpoint ~5.55e-17mm outside the pad. Downstream DRC (`isPointInPad`) then reports a false "disconnected endpoint" error.


seveibar (changes requested):
no tsx = no repro

itisrohit:
This repro belongs in @tscircuit/core or @tscircuit/checks — the footprinter geometry is correct. The bug is isPointInPad missing an epsilon tolerance.

PR #754, opened 2026-07-30
feat(stampboard): support offset side rows and inner SMT grids

techmannih:
## Summary

This PR extends the `stampboard` footprint generator to support footprints containing offset side rows and an internal SMT pad grid.

### Changes

- Adds `sidey` for a shared Y offset on the left and right pad rows.
- Adds `innergrid` to define the internal SMT pad grid dimensions.
- Adds `innerp` for the internal grid pitch.
- Adds `innerpw` and `innerph` for internal pad dimensions.
- Adds `innerx` and `innery` for the internal grid center position.
- Includes internal pads in total pad numbering.
- Exposes the new parameters through the Footprinter builder API.
- Adds geometry assertions and an SVG snapshot test.

All new parameters are opt-in, so existing `stampboard` footprints remain unchanged.

## Motivation

JLCPCB part `C2913202` (`ESP32-S3-WROOM-1-N16R8`) contains:

- 14 left-side SMT pads
- 14 right-side SMT pads
- 12 bottom SMT pads
- A 3×3 internal SMT pad grid
- 49 pads total

The existing `stampboard` generator could represent the perimeter layout, but it could not shift the side rows or generate the internal SMT grid.

Because of this, the current automatic discovery selected an incorrect `qfn49` footprint with very low copper overlap.

## C2913202 Benchmark

The benchmark used the exact `C2913202` footprint fetched from EasyEDA and compared it using `fast-footprint-compare`.

### Results

| Result | Copper IoU | Pad Count |
|---|---:|---:|
| Before: current automatic `qfn49` fallback | 3.5742% | 49 vs 49 |
| After: new `stampboard` geometry | 99.8854% | 49 vs 49 |

**Improvement: +96.3112 percentage points**

```text
stampboard_left14_right14_bottom12_top0_w19.000089mm_h18.9899036mm_p1.27mm_pw0.8999982mm_pl1.499997mm_sidey0.7899527mm_innergrid3x3_innerp1.400048mm_innerpw0.8999982mm_innerph0.8999982mm_innerx-1.500124mm_innery1.3248767mm
```

Copper IoU:

```text
99.8854%
```


src/fn/stampboard.ts
```diff
@@ -22,6 +23,13 @@ export const stampboard_def = base_def.extend({
   p: length.default(length.parse("2.54mm")),
   pw: length.default(length.parse("1.6mm")),
   pl: length.default(length.parse("2.4mm")),
+  sidey: length.default(0).describe("shared Y offset for left and right rows"),
```
seveibar: this is unclear, needs a much better api

```tsx
  sidey: length.default(0).describe("shared Y offset for left and right rows"),
```

seveibar (commented):
needs better/more clear api

src/fn/stampboard.ts
```diff
@@ -22,26 +23,36 @@ export const stampboard_def = base_def.extend({
   p: length.default(length.parse("2.54mm")),
   pw: length.default(length.parse("1.6mm")),
   pl: length.default(length.parse("2.4mm")),
+  leftrowy: length.default(0).describe("left pad row center y position"),
+  rightrowy: length.default(0).describe("right pad row center y position"),
```
seveibar: this still isn't good, we never use "centerx" or "centery" as an anchoring position unless it is the true center of a symmetric piece

seveibar (changes requested):
see comments, lots of forbidden terms

seveibar (commented):
Offset from what? This isnt clear

seveibar:
This isnt clear and isnt associated with a standard. The offsets would need to be proposed in a separate pr than the “inner grid” but the inner grid isnt something we would support- its not that conventional

techmannih:
damm

PR #760, opened 2026-08-01
fix(bga): use circular pads by default for KiCad parity

anil08607:


tests/kicad-parity/bga165_kicad_parity.test.ts
```diff
@@ -17,9 +17,14 @@ test("parity/bga165", async () => {
...
+      .filter((element) => element.type === "pcb_smtpad")
+      .every((pad) => pad.shape === "circle"),
+  ).toBe(true)
   expect(svgContent).toMatchSvgSnapshot(import.meta.path, "bga165")
   expect(booleanDifferenceSvg).toMatchSvgSnapshot(
     import.meta.path,
     "bga165_boolean_difference",
   )
-})
+}, 10_000)
```
techmannih: I think we don't need to add the timeout

techmannih:
Also try to fix the diff percentage. Footprint should be match completely

tests/kicad-parity/bga165_kicad_parity.test.ts
techmannih: ??? @anil08607

tests/kicad-parity/__snapshots__/bga165.snap.svg
techmannih, line 1: There is also same courtyard issue no?

anil08607: yes, repro has been added for that issue.

techmannih (approved):

src/fn/bga.ts
```diff
@@ -77,7 +81,9 @@ export const bga = (
 
   ball ??= (0.75 / 1.27) * p
 
-  pad ??= ball * 0.8
+  // KiCad BGA land patterns use a pad diameter equal to half the ball pitch.
+  // Preserve the existing ball-based derivation when the ball is overridden.
```
seveibar: Do not reference kicad in the source code. We do NOT copy kicad, we implement IPC standards. Never reference kicad in the source code because it misleads developers and ai

remove the comment

seveibar (commented):
See comment

anil08607:
@seveibar Done!!

seveibar (approved):
this may cause upstream issues for calculate-packing and the autorouter that we will want to be on top of due to the curved surfaces

PR #767, opened 2026-08-04
fix(soic): match KiCad silkscreen and courtyard parity

KrishnaX12:
Fixes SOIC courtyard to match KiCad parity (IoU < 1% diff) by using precise pinRowSpanY/bodyHalfHeight calculations instead of Math.max/min guessing. Adds bw/bh body dimension support and a KiCad-standard Pin 1 silkscreen chevron marker.


techmannih (commented):
We are not copy the silkscreen from kicad

PR #769, opened 2026-08-04
Support asymmetric left/right pad sizes in QFN footprints

Abse2001:
## Motivation

Some QFN packages use different pad dimensions on their left/right sides than on their top/bottom sides.

C128396 (`TS3USB221ARSER`) uses:

- Top/bottom pads: `0.25 × 0.525 mm`
- Left/right pads: `0.30 × 0.58 mm`

The existing QFN implementation required every side to share `pw` and `pl`, limiting its best IoU to approximately 93%.

## Implementation

- Add `lrpw` and `lrpl` parameters for left/right pad dimensions.
- Keep `pw` and `pl` as the defaults when the overrides are omitted.
- Use the overrides when positioning pads and calculating thermal-pad and courtyard geometry.
- Add a C128396 regression test and SVG snapshot.

## Validation

- `bun run build` passes.
- All 517 Footprinter tests pass.
- C128396 can now be represented at 99.9564% copper IoU.
-


Abse2001:
@seveibar

PR #771, opened 2026-08-07
Support asymmetric SOT-343 pad geometry

Abse2001:
## Summary

- extend the existing sot343 family instead of adding a BFP650-specific function
- add measured row span, pin 2 pad length, and pin 2 center-offset parameters
- preserve the standard SOT-343 defaults and KiCad parity
- add a C151520 regression test and SVG snapshot

## Motivation

C151520 is officially packaged as SOT-343, but its recommended land pattern uses a wider, inward-shifted pin 2 pad. This geometry should be represented through the existing standard family rather than a manufacturer-specific footprint name.

## Implementation

The existing sot343 generator now supports rowspan, pin2padlength, and pin2padcenteroffsetx. Every new parameter is exercised by the real C151520 regression. Full pin-indexed parameter names are accepted without changing compound dimensions such as grid3x3 or thermalpad2x2.4mm.

## Validation

- C151520 copper IoU: 99.978%
- standard SOT-343 KiCad courtyard parity: 100.00%
- 519 Footprinter tests passed
- focused parser and SOT-343 tests passed
- bun run build passed
- SVG snapshot added for the parameterized C151520 footprint


seveibar (changes requested):
Isnt that a manufacturer name?

Abse2001:
Yes, BFP650 is an Infineon product name. I checked the official documentation: the device package is SOT343, which Footprinter already supports, while the asymmetric recommended land pattern introduced here is not a separately standardized package. This PR is therefore manufacturer-specific, so I am closing it.

Abse2001:
Follow-up: the geometry can be supported without a manufacturer-specific family. Replacement PR #780 removes sot343bfp650 and adds measured parameters to the existing sot343 generator, with C151520 regression coverage and unchanged standard parity.

PR #772, opened 2026-08-07
Add LED3510 footprint

Abse2001:
## Summary

- add a dedicated `led3510` four-contact inline LED footprint
- expose only the measured outer/inner pad geometry used by this package
- add a C22461789 regression test and SVG snapshot

## Motivation

XL-3510RGBA-SG (JLCPCB C22461789) has four pads in one row, with wider outer contacts and a slightly offset inner pair. Existing two-pad LED families cannot represent that topology; the best unrelated FPC match reached only 89.08% copper IoU.

## Implementation

The new explicit `led3510` variant models the two outer contacts and the inner pair using package-level dimensions rather than arbitrary per-pad definitions. Its defaults are measured from C22461789.

## Validation

- `bun test tests/led3510.test.ts`: 1 passed
- `bun run build`
- C22461789 copper IoU: 99.989%
- new SVG snapshot for the real land pattern


seveibar (changes requested):
Not a standard

Abse2001:
I could not establish LED3510 as a standardized footprint. The 3510 designation describes the body size, while this four-contact inline land pattern was measured specifically from XL-3510RGBA-SG. It should not become a general Footprinter family, so I am closing this PR.

PR #773, opened 2026-08-07
Add SMD potentiometer footprint

Abse2001:
## Summary

- add an explicit `smdpotentiometer` footprint family
- model the wider center wiper pad and equal outer terminal pads
- add a C78322 regression test and SVG snapshot

## Motivation

3RL090M-5-S (JLCPCB C78322) is a three-pad SMD potentiometer. The existing `potentiometer` family is through-hole, so discovery fell back to `solderjumper3` and reached only 88.34% copper IoU.

## Implementation

The new family uses a shared terminal pitch, measured outer-pad width, center-pad width, and pad height. These parameters describe the real potentiometer package without exposing arbitrary per-pad placement.

## Validation

- `bun test tests/smdpotentiometer.test.ts`: 1 passed
- `bun run build`
- C78322 copper IoU: 99.998%
- new SVG snapshot for the real land pattern


seveibar (changes requested):
Im nit sure how standard this is for smd potentiometers?

Abse2001:
I rechecked C78322 and found that it was misclassified: 3RL090M-5-S is a Yageo three-electrode gas-discharge tube, not a potentiometer. The proposed geometry belongs to the manufacturer 3R-5-S product series and is not a standard SMD potentiometer footprint. I am closing this PR.

Abse2001:
Replaced by #781. C78322 was misclassified here: the 3RL090M-5-S is a three-electrode SMD gas discharge tube, not a potentiometer. The replacement models the reusable cross-vendor package class as `smdgdt3`, with every parameter backed by the C78322 land pattern.

PR #776, opened 2026-08-07
Add G5NB relay footprint

Abse2001:
## Summary

- add an explicit `relayg5nb` footprint variant
- model its two coil holes and two differently sized contact holes
- add a C225714 regression test and SVG snapshot

## Motivation

G5NB-1A-E DC12 (JLCPCB C225714) has a sparse, nonuniform four-hole layout. Modeling it as a nominal 2x9 pin grid lost the two hole-size groups and reached only 87.27% copper IoU.

## Implementation

The package-specific variant encodes the documented G5NB land pattern directly. It intentionally exposes no speculative per-pin coordinate parameters.

## Validation

- `bun test tests/relayg5nb.test.ts`: 1 passed
- `bun run build`
- C225714 copper IoU: 99.993%
- new SVG snapshot for the real relay footprint


seveibar (changes requested):
Not a standard

Abse2001:
Confirmed: G5NB is an Omron relay product series, and the mounting-hole arrangement in this PR is manufacturer-specific rather than a standardized relay package. It should not be added as a public Footprinter family, so I am closing this PR.

PR #779, opened 2026-08-07
Add UFQFPN-20 footprint

Abse2001:
## Summary

- add an explicit `ufqfpn20` footprint variant
- preserve its eight chamfered corner leads as polygon copper
- add a C35556 regression test and SVG snapshot

## Motivation

STM8S003F3U6TR (JLCPCB C35556) uses a UFQFPN-20 land pattern with twelve rectangular leads and eight chamfered corner leads. Symmetric LGA/QFN pads cannot reproduce the corner copper, limiting the best match to 85.75% IoU.

## Implementation

The package-specific UFQFPN-20 variant encodes the verified standard land pattern and its polygon corners. It exposes no speculative polygon or per-pad controls.

## Validation

- `bun test tests/ufqfpn.test.ts`: 1 passed
- `bun run build`
- C35556 copper IoU: 100.00%
- new SVG snapshot showing all eight chamfered corner leads


seveibar (changes requested):
idgi, footprint doesnt support pin counts, is this an ipc standard

PR #781, opened 2026-08-07
Add a generic linear SMD pad footprint

Abse2001:
## Summary

- add a neutral `smdpads` family for an evenly pitched linear row of SMD pads
- support an optional wider geometric center pad on odd pad counts
- add real regression fixtures for C78322 and C7464756
- avoid both a component-specific GDT family and the semantically incorrect `pinrow` model

## Motivation

A human would describe this copper geometry as three evenly pitched SMD pads with a wider center pad. It is not a connector pin row, and it is not specific to GDTs or one manufacturer: C78322 and the unrelated C7464756 LED use the same reusable geometry at different dimensions.

The footprint is therefore represented by the explicit geometric name `smdpads3`, following the review suggestion.

## Implementation

The family uses:

- `p`: pad center pitch
- `pw`: outer pad width
- `ph`: pad height
- `centerpadwidth`: optional geometric center-pad width

`centerpadwidth` is accepted only for odd pad counts, where a single geometric center pad exists.

C78322 is represented by:

```
smdpads3_p3.299968mm_pw1.7999964mm_ph5.7999884mm_centerpadwidth2.3999952mm_rounded0
```

C7464756 is represented by:

```
smdpads3_p1.299972mm_pw0.3999992mm_ph3.0999938mm_centerpadwidth1.1999976mm_rounded0
```

## Validation

- C78322 copper IoU: **100.00%**
- C7464756 copper IoU: **100.00%**
- `bun test`: **521 pass, 0 fail**
- `bun run build`: pass
- new C78322 SVG regression snapshot


seveibar (changes requested):
this is a bit rough, it's not really a pinrow so not semantically correct

try something like `smdpads3...` maybe? Think "how would a human describe this"

seveibar (approved):

PR #783, opened 2026-08-08
fix: default vson dimensions so vson8 renders from a bare name

### LINKED ISSUE #782: vson8 throws a raw parse error while son8/wson8 render from a bare name
`vson` is listed by `getFootprintNames()`, but calling it from a bare name crashes:

```ts
fp.string("vson8").circuitJson() // throws a raw ZodError (Required: p, w, grid, pinw, pinh)
fp.string("son8").circuitJson()  // 8 pads
fp.string("wson8").circuitJson() // 9 pads
```

Both sibling SON-family footprints (`son`, `wson`) default their dimensions and render from a bare name. `vson` alone declares `p/w/grid/pinw/pinh` with no defaults (the def carries a comment "can't use defaults because there is not a lot of common dimensions"), so any bare `vsonN` throws. A user who picks `vson8` from the footprint list gets a raw parse error instead of a footprint.

zkasuran:
Closes #782

`vson` is advertised by `getFootprintNames()`, but `fp.string("vson8").circuitJson()` throws a raw ZodError because `vson_def` declares `p`, `w`, `grid`, `pinw` and `pinh` with no defaults. The sibling SON-family defs (`son_def`, `wson_def`) default every dimension and render from a bare name, so `son8` and `wson8` work while `vson8` crashes.

## Repro

```ts
fp.string("vson8").circuitJson() // throws on main
fp.string("son8").circuitJson()  // 8 pads
```

## The change

Give `vson_def` defaults that describe the canonical KiCad VSON-8-1EP 3x3mm 0.65mm-pitch part (`Package_SON.pretty/VSON-8-1EP_3x3mm_P0.65mm_EP1.65x2.4mm`): pitch 0.65mm, row spacing 2.9mm, 3x3mm body, 0.85mm by 0.35mm pads. A bare `vson8` now renders 8 pads like `son8` and `wson8`. Its courtyard also lines up with that real KiCad footprint. Explicit dimensions still override for a specific part. The exposed pad default stays 0 so a bare `vson8` keeps 8 pads; pass `ep` for a part that needs the thermal pad. The old "can't use defaults" comment is dropped, since son and wson show the SON family does have workable defaults.

## Verification

- `tests/vson8.test.ts` renders `vson8` to a PCB SVG snapshot and asserts 8 pads.
- `tests/kicad-parity/vson8_kicad_parity.test.ts` adds a parity case for the bare `vson8` against the KiCad VSON-8-1EP reference. Courtyard diff is 0.19%, under the 0.5 threshold the other SON parity tests use.
- Full suite green (523 pass), `bun run build` clean, `biome format` clean.

## Disclosure

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: the full `bun test` suite (523 pass), `bun run build` and `biome format`.


techmannih (commented):
Please add kicad parity test

zkasuran:
Added the kicad parity test. It compares the bare `vson8` (the new defaults) against the canonical KiCad VSON-8-1EP 3x3mm 0.65mm-pitch part, `Package_SON.pretty/VSON-8-1EP_3x3mm_P0.65mm_EP1.65x2.4mm`. It lives in `tests/kicad-parity/vson8_kicad_parity.test.ts` next to the existing explicit-dims case.

To make the defaults match a real part I retuned them to that VSON-8: pitch 0.65mm, row spacing 2.9mm, 3x3mm body, 0.85mm by 0.35mm pads. Courtyard diff comes out at 0.19%, under the 0.5 threshold the other SON parity tests assert. The 8 pins are dimensionally identical to KiCad's.

I kept the exposed pad at 0 so a bare `vson8` stays 8 pads and `tests/vson8.test.ts` remains valid. That is the one thing that differs from the KiCad part, which carries a 1.65x2.4mm thermal pad, so the pad-area label in the overlay reads high while the pins and courtyard match. Happy to fold the EP into the default if you would rather the bare name include the thermal pad.

Full suite is green (523 pass), `bun run build` and `biome format` clean.

imrishabh18 (approved):

tests/kicad-parity/__snapshots__/vson8_default.snap.svg
techmannih, line 1: why are we missing thermal pad, is it default?

src/fn/vson.ts
```diff
@@ -12,21 +12,29 @@ import { dim2d } from "src/helpers/zod/dim-2d"
 import { type SilkscreenRef, silkscreenRef } from "src/helpers/silkscreenRef"
 import { createRectUnionOutline } from "src/helpers/rect-union-outline"
 
-// can't use defaults because there is not a lot of common dimensions.
```
techmannih, line 15: ?????

tests/kicad-parity/__snapshots__/vson8_default.snap.svg
zkasuran: The thermal pad is deliberately opt-in via `ep`, not part of the bare default. Its size is part-specific, so a fixed default overflows smaller VSON bodies. Concretely, defaulting `ep` to the 3x3mm part's 1.65x2.4mm makes the existing `VSON8_grid1.5x2mm` test render a pad bigger than its own 1.5x2mm body.

So a bare `vson8` renders the 8 signal pins plus the matching courtyard; the parity sits at 0.19% because the EP falls inside the courtyard. Pass `ep=1.65x2.4mm` to add the thermal pad for the full 1EP part. This matches the sibling `son8`, which also defaults its EP off.

If you would prefer the bare `vson8` to be the full VSON-8-1EP, I can default the EP on and scale it to the body so smaller parts do not overflow. Your call.

src/fn/vson.ts
zkasuran: These defaults are one specific real part, the canonical KiCad VSON-8-1EP 3x3mm 0.65mm-pitch (`Package_SON/VSON-8-1EP_3x3mm_P0.65mm_EP1.65x2.4mm`), not a claim that every VSON shares those dims. The old comment was right that VSON dims vary, so instead of inventing an average I anchored the bare-name default to that one canonical part and pinned it with a KiCad parity test (courtyard 0.19%).

A bare `vson8` now renders that real part instead of throwing; any other VSON is still just a matter of passing explicit dims. Same pattern the other defaulted footprint fns use.

PR #785, opened 2026-08-08
fix(solderjumper): default to 2 pins when no pin count is given

### LINKED ISSUE #784: solderjumper without a pin count emits NaN geometry
### Summary

### Reproduction

```ts

```

Output at v0.0.417:

- 0 pads

### Cause

### Expected

zkasuran:
Closes #784

## What

A bare `fp.string("solderjumper")` (no pin count) silently produced invalid geometry: a `pcb_courtyard_rect` with `width` and `center.x` set to NaN plus a `pcb_silkscreen_text` with a NaN anchor. `solderjumper2` and `solderjumper3` were already fine.

The cause is in `src/fn/solderjumper.ts`: `num_pins` was read from params with no default, so an omitted count left it `undefined` and `(num_pins - 1) * padSpacing` (and everything derived from it) became NaN.

## Fix

Default `num_pins` to 2 when it is omitted, the same way the function already defaults `p`, `pw` and `ph`. This matches the wider convention in this repo where a missing pin count falls back to a sensible default (`dip` gives 6, `ssop` gives 8, `pinrow` gives 6, plus many others). 2 lines up with `solderjumper2`, the smallest documented variant and the first example in the doc comment. Explicit counts are unchanged.

## Tests

- An assertion test: a bare `solderjumper` now yields 2 pads with finite courtyard geometry and a finite silkscreen text anchor (no NaN). It fails on the current code (0 pads, NaN width) then passes with the fix.
- A visual SVG snapshot (`solderjumper_default`) rendered from the fixed output, showing the real 2-pin footprint.

## Verification

Verified locally before submitting:

- `bun test`: 523 pass, 0 fail
- `bun run build`: success (ESM plus dts)
- `bunx biome format` clean on the changed files
- confirmed the assertion goes red on the pre-fix code then green after the fix. Opened the generated snapshot and confirmed it shows a real footprint

## AI disclosure

AI assistance (Claude, Anthropic) was used in developing this change. Verified locally before submitting: full `bun test` (523 pass, 0 fail), plus `bun run build` and `biome format` on the changed files.


zkasuran:
Added a KiCad parity test up front so this does not need a separate round.

`tests/kicad-parity/solderjumper_kicad_parity.test.ts` now has a `parity/solderjumper-default-pincount` case. It renders the bare `solderjumper` form this PR enables (no pin count, so `num_pins` falls back to the new default of 2) and compares it against the real KiCad `SolderJumper-2_P1.3mm_Open_Pad1.0x1.5mm` footprint. Courtyard diff is 0.00%, well under the 0.5 threshold the other parity tests use, so the defaulted output is not just the right pad count, it is geometrically the correct SolderJumper-2.

One honest note on dimensions. footprinter's own solderjumper defaults (2.54mm pitch, 1.5mm square pads) are a generic 0.1 inch jumper that does not track any KiCad standard part. KiCad's SolderJumper-2 is a smaller 1.3mm-pitch bridge pad. So the parity case supplies KiCad's dimensions explicitly (`solderjumper_p1.3_pw1_ph1.5`), the same way the existing `solderjumper2` parity case does. What it proves is that the new default-pin-count path produces the same KiCad-correct geometry the explicit form already did. The pure 2.54mm default is left covered by the finite-geometry assertion plus the `solderjumper_default` snapshot already in this PR.

Full suite green: 524 pass, 0 fail. `bun run build` clean, `biome format` clean on the changed file.

PR #787, opened 2026-08-08
fix(fpc): default pin count to 12 so a bare fpc renders (#786)

### LINKED ISSUE #786: fpc throws a raw parse error when called by its bare name, unlike every other pin-count footprint
## What happens

```ts

```

```
ZodError: [
  {
    "code": "invalid_type",
    "expected": "number",
    "received": "nan",
    "path": ["num_pins"],
  }
]
```

## Why it happens

```ts
```

## Why it is a bug

- `son` to 8, `sot223` to 4

## Expected

zkasuran:
## Summary

- default `fpc`'s pin count to 12 so the bare `fpc` name renders instead of throwing a raw Zod parse error
- add a regression test that asserts bare `fpc` produces the 12-pin FPC-05F-12PH20 layout, with an SVG snapshot

Closes #786

## Motivation

`fp.string("fpc").circuitJson()` threw a `ZodError` on `num_pins` because the schema had no default. Every other pin-count footprint defaults its count (bga 64, quad 64, ssop 8, dip 6, lga 14, son 8, sot223 4, the two-pin parts 2), so `fpc` was the only one that failed on a bare name.

The `fpc` dimensional defaults are already built around a 12-pin part. The existing `fpc12` test reproduces FPC-05F-12PH20 from the schema defaults alone, so 12 is the count the rest of the schema was tuned for.

## Implementation

One line in `src/fn/fpc.ts`:

```ts
num_pins: z.coerce.number().int().min(2).default(12),
```

The default only applies when no count is given, so `fpc12`, `fpc30_staggered` and every explicit form stay unchanged. Bare `fpc` now resolves to the same geometry as the explicit `fpc12` reference.

## On KiCad parity

The part this footprint models, FPC-05F-12PH20, is not in the KiCad cache. The nearest 12-pin 0.5mm entry there is a different manufacturer part, Hirose FH12-12S-0.5SH, whose mounting pads sit at ±4.65 (ours at ±4.44) with a different body origin. A parity run against it gives a 41% courtyard difference. Matching it would distort the FPC-05F-12PH20 geometry the repo already asserts. So this test follows the existing fpc test style: a hardcoded copper-pattern check against the reference part plus an SVG snapshot, not a kicad-parity test.

## Validation

- `bun test`: 522 pass, 0 fail
- `bun run build`: success
- `biome format`: clean
- new test: bare `fpc` renders 14 pads matching the explicit `fpc12` (FPC-05F-12PH20) reference, plus a new SVG snapshot

## AI disclosure

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: `bun test` (522 pass, 0 fail), `bun run build` and `biome format`.


techmannih (commented):
kicad test?

zkasuran:
Added one, though it needs a caveat because `fpc` does not map cleanly onto the usual kicad-parity check.

KiCad has no single canonical FPC footprint. Every part in `Connector_FFC-FPC.pretty` is a specific vendor connector (Hirose, Amphenol, TE ...) with its own housing, mounting-pad geometry and courtyard. The footprinter default reproduces the FPC-05F-12PH20 land pattern, which is not in the KiCad library at all. What every 0.5mm 12-pin FFC part does share is the contact land pattern, so that is what I pinned.

`tests/kicad-parity/fpc_kicad_parity.test.ts` compares the bare `fpc` default against Hirose FH12-12S-0.5SH:
- 12 contacts, matching KiCad in count, 0.5mm pitch, 0.3mm pad width and X positions (-2.75mm to 2.75mm).
- an overlay snapshot of footprinter against the KiCad reference.

It deliberately does not assert courtyard parity, unlike the sibling tests. Footprinter draws a pad-bounding-box courtyard (11.38 x 4.95mm); the KiCad courtyard encloses the connector body (12.12 x 7.9mm on the Hirose part). Measured courtyard IoU against the four nearest KiCad 12-pin 0.5mm parts (Hirose FH12-12S-0.5SH, Amphenol F32Q/F32R-1A7x1-11012, TE 1-1734839-2) is 47 to 67% (diff 34 to 53%), all of it the housing, which footprinter does not model and which differs by vendor. The other parity tests assert well under 5%, so a courtyard bound here would either fail or be a made-up number that certifies nothing.

Verified locally: `bun test tests/kicad-parity/fpc_kicad_parity.test.ts` passes (30 assertions) and `biome format` is clean on `src` and `tests`.

If you would rather point it at a different reference part or record the courtyard as a documented non-gating number, happy to adjust.

PR #795, opened 2026-08-10
fix(lcc): default to 1.27mm LCC/PLCC pitch so pads stop overlapping

### LINKED ISSUE #794: lcc footprints default to a 0.5mm pitch and emit overlapping pads

Repro:

```ts

```

zkasuran:
Closes #794

## What changed

lcc set pl and pw but never p, so it fell back to the generic quad default pitch of 0.5mm. With 0.6mm-wide pads on a 0.5mm pitch, neighboring pins overlap by 0.1mm, which is a short in the emitted PCB. Every bare lcc size is affected (lcc20, lcc28, lcc44, lcc68).

LCC and PLCC are a 1.27mm pitch family (JEDEC MO-047 / MS-034). The repo already uses that pitch in the lcc68 parity test against PLCC-68_24.2x24.2mm_P1.27mm. This PR defaults the lcc pitch to 1.27mm when the caller gives none, in the same block that already defaults pl and pw. An explicit pitch still wins, so lcc68_..._p1.27mm and any custom pitch are unchanged.

## Repro (before this PR)

```ts
// pin 1 at (-3.150, 2.500), pin 2 at (-3.150, 2.000), both 1 x 0.6
// centers 0.5mm apart, pads 0.6mm tall  ->  0.1mm overlap
```

After the fix the same pads sit 1.27mm apart with 0.67mm of clearance.

## Tests

- tests/lcc.test.ts: a new assertion that no two pads in lcc20, lcc44 or lcc68 overlap. It fails before the fix and passes after.
- tests/kicad-parity/lcc44_kicad_parity.test.ts: a KiCad parity test for lcc44 against PLCC-44. Pad area matches within 0.00% and the courtyard is within 4.02%.
- An SVG snapshot of the default lcc44.

## Verification

Run locally on this branch:
- bun test: 524 pass, 0 fail across 224 files
- bunx biome format .: clean, 384 files
- bun run build: success (esm and dts)

AI assistance (Claude) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: bun test (524 pass, 0 fail), bunx biome format . (clean), bun run build (success).


zkasuran:
Closing this. Refocusing on building real boards rather than footprint tweaks. Thanks for the steer.

PR #799, opened 2026-08-14
fix(passive): define courtyards for dimensioned footprints

AnasSarkiz:
## Summary

- require each passive component function (`res`, `cap`, `diode`, and `led`) to define its courtyard dimensions
- build dimensioned courtyard envelopes from copper, explicit body dimensions, and the generated silkscreen outline
- apply the repository-standard 0.25 mm assembly clearance per side
- preserve all predefined StandardSize courtyard dimensions and existing pad geometry

## Root cause

`passive()` previously emitted a courtyard only when a metric or imperial StandardSize supplied explicit dimensions. Fully dimensioned footprint strings therefore returned no `pcb_courtyard_*` record.

## Review follow-up

The original revision put a generic sizing policy inside `passive()` and only considered copper/body geometry. That produced snapshots where the courtyard crossed the generated silkscreen outline.

The revised implementation moves courtyard ownership to each component function. `passive()` supplies resolved geometry context, while each component explicitly defines its dimensions. Dimensioned envelopes now contain copper, body, and silkscreen geometry before assembly clearance is applied.

## Tests

Focused geometry and SVG snapshot coverage includes:

- `res_p0.8656mm_pw0.5657mm_ph0.54mm`
- `res_p1.5067mm_pw0.8065mm_ph0.864mm`
- generic capacitor, diode, and LED footprints
- a passive with explicit body width and height
- a THT passive
- predefined `res0402` and `res0603` footprints

The tests assert exactly one courtyard, finite positive dimensions, copper and silkscreen containment, 0.25 mm clearance, unchanged pad/hole geometry and port hints, and unchanged predefined courtyard dimensions. Updated SVG snapshots were visually inspected.

## Validation

- `bun test` — 530 passed, 0 failed
- focused courtyard/resistor tests — 22 passed, 0 failed
- `bun run build`
- `bun run format`

## Consumer impact

After release, the MangoPi board can be rebuilt to verify courtyard coverage increases from 20/137 to 135/137. TP1 and TP2 are expected to remain courtyard-less because they are built-in tscircuit testpoints requiring a separate core fix.


seveibar (changes requested):
define courtyards properly! anytime you have "fallback" anywhere it should basically be an insta-reject. Each function should define it's courtyard- no hacks!

seveibar (changes requested):
even worse there are clearly buggy courtyards in the snapshots. boooo!

seveibar (changes requested):
boooo

PR #818, opened 2026-08-21
feat: support explicit courtyard dimensions for custom passives

MustafaMulla29:
## Summary

- add cyw and cyh parameters for custom resistor, capacitor, diode, LED, and smdpads footprints
- use those values as the exact courtyard width and height
- preserve existing standard-package courtyards and smdpads fallback behavior
- include courtyard-visible SVG snapshots for every supported custom family

## Validation

- bun test: 561 passing
- bun run build: passing


seveibar (approved):
I dont think you fixed anything, iiuc you just added an alias

seveibar (approved):

PR #823, opened 2026-08-23
Fix footprint defaults: tqfp44, lcc, to220 pitch + pad/fpc/solderjumper bare-name handling

### LINKED ISSUE #784: solderjumper without a pin count emits NaN geometry

### LINKED ISSUE #786: fpc throws a raw parse error when called by its bare name, unlike every other pin-count footprint

### LINKED ISSUE #788: pad throws a raw TypeError from a bare name while smtpad defaults its dimensions
`pad` is listed by `getFootprintNames()`, but calling it from a bare name crashes:

```ts
fp.string("pad").circuitJson()    // throws TypeError: undefined is not an object (evaluating 'n.replace')
fp.string("smtpad").circuitJson() // renders a 1mm pad
```

`pad_def` declares `w` and `h` with no defaults; `pad()` then reads them straight off the params and calls `mm(w)` without parsing, so a missing dimension reaches `@tscircuit/mm` as `undefined` and throws a low-level `TypeError`. The sibling `smtpad` handles the same case by defaulting to a 1mm pad (`else width = mm("1mm")` in `smtpad_def`), so `smtpad` renders from a bare name while `pad` crashes. A user who picks `pad` from the footprint list gets a raw `TypeError` that leaks the `mm()` internals instead of a footprint.

### LINKED ISSUE #790: to220 uses a body-width-derived pin pitch (2.6mm) instead of the standard 2.54mm

### LINKED ISSUE #792: tqfp44 generates a 0.5mm pitch with 0.30mm pads, but TQFP-44 is a 0.8mm-pitch part
### What happens

`fp.string("tqfp44")` (and `tqfp44_w10`) produces a 44-pin QFP on a 0.5mm pitch with 0.30mm-wide pads.

### What it should be

TQFP-44 is standardized as a 10x10mm body on 0.8mm pitch (JEDEC MS-026 BBA). KiCad's `TQFP-44_10x10mm_P0.8mm` uses 0.8mm pitch with 1.475 x 0.55mm pads. footprinter already emits exactly those values for its sibling TQFP-32, which is also a 0.8mm-pitch part. The current output matches no real TQFP-44, so a board that uses `tqfp44` gets pads at 0.5mm pitch instead of 0.8mm and too narrow to sit under the leads.

### Root cause

In `src/fn/tqfp.ts` the pitch and pad-width defaults place `case 44` in the 0.5mm-pitch, 0.30mm-pad group with 48/64/100. TQFP-44 belongs with `case 32` (0.8mm, 0.55mm).

### Evidence

| | footprinter `tqfp44` | KiCad TQFP-44_10x10 |
|---|---|---|
| pitch | 0.5mm | 0.8mm |
| pad width | 0.30mm | 0.55mm |
| pad length | 1.475mm | 1.475mm |

I have a fix ready with a kicad-parity test and will open a PR.

### LINKED ISSUE #794: lcc footprints default to a 0.5mm pitch and emit overlapping pads

Luvi-1:
/claim #792

Fixes #792
Fixes #794
Fixes #790
Fixes #788
Fixes #786
Fixes #784

### tqfp44 (#792)
Moved from 0.5mm to 0.8mm pitch and 0.55mm pad width to match JEDEC MS-026 BBA / KiCad `TQFP-44_10x10mm_P0.8mm`.

### lcc (#794)
Defaulted pitch to 1.27mm (JEDEC MO-047 / MS-034 family standard). Previously inherited 0.5mm from quad default, causing 0.6mm pads on 0.5mm pitch (overlapping by 0.1mm).

### to220 (#790)
Replaced body-width-derived pitch with fixed JEDEC 2.54mm lead pitch. Previously computed 2.6mm for 3-pin parts (0.06mm off). The `p` parameter was declared but never read — now honored.

### pad (#788)
Defaulted w/h to 1mm so bare `fp.string("pad")` renders instead of TypeError.

### fpc (#786)
Defaulted num_pins to 12 so bare `fp.string("fpc")` renders the FPC-05F-12PH20 footprint instead of ZodError.

### solderjumper (#784)
Defaulted num_pins to 2 so bare `fp.string("solderjumper")` emits valid geometry instead of NaN courtyard/silkscreen.

## Testing

All **556 tests pass**, zero failures:
- KiCad parity tests all green including TO-220F-3 (courtyard IoU 99.83%)
- Updated SVG snapshots for to220_2–5 and tqfp44_w10


Luvi-1:
Friendly ping @seveibar — this fixes 6 open bugs (#792 #794 #790 #788 #786 #784) with all CI passing. Happy to address any feedback.

PR #824, opened 2026-08-23
align JST PH footprints with KiCad pad geometry

anil08607:
## Summary

- Update JST-PH defaults to use 2.00 mm pitch and 0.75 mm drill holes
- Match KiCad pad geometry:
  - Rounded rectangular pad for pin 1
  - Oval pads for remaining pins
- Align JST-PH courtyards with KiCad
- Add KiCad parity tests and SVG snapshots for 2, 4, and 6-pin variants
- Keep the existing simplified silkscreen

## Testing

- 18 focused tests pass
- All parity snapshots report 0.00% pad difference
- Courtyard overlap is 100%
- `bun run build` passes


techmannih (commented):
your silkscreen is static for all variants? that should be dynamic for all variants

techmannih (approved):

PR #825, opened 2026-08-25
fix(solderjumper): default num_pins to 2 when omitted to prevent NaN geometry

### LINKED ISSUE #784: solderjumper without a pin count emits NaN geometry

moxiaoshen567:
Fixes #784

## Summary
- Defaults `num_pins` to `2` in `src/fn/solderjumper.ts` so calling `fp.string("solderjumper")` without an explicit pin count renders a valid 2-pin footprint instead of emitting `NaN` geometry / 0 pads.
- Added unit test in `tests/solderjumper.test.ts` verifying bare string call produces 2 pads with finite courtyard and silkscreen coordinates.

## Verification
```bash
bun test tests/solderjumper.test.ts
# 6 pass, 0 fail
```

/claim #784

### DIFF
--- a/src/fn/solderjumper.ts
+++ b/src/fn/solderjumper.ts
@@ -20,13 +20,13 @@ import { length } from "circuit-json"
  *   solderjumper({ num_pins: 3, pw: 2.0, ph: 1.0 }) // custom pad size
  */
 export const solderjumper = (params: {
-  num_pins: 2 | 3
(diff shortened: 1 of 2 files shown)


moxiaoshen567:
### 💰 Payout Details

PR #826, opened 2026-08-25
fix(fpc): default num_pins to 12 when omitted to prevent parse error on bare name

### LINKED ISSUE #786: fpc throws a raw parse error when called by its bare name, unlike every other pin-count footprint

moxiaoshen567:
Fixes #786

## Summary
- Defaults `num_pins` to `12` in `fpc_def` (`src/fn/fpc.ts`) so calling `fp.string("fpc")` without an explicit pin count renders the standard 12-pin FPC-05F-12PH20 footprint (14 pads) instead of failing Zod coercion with `NaN`.
- Added unit test in `tests/fpc.test.ts` verifying bare string call produces 14 pads without throwing.

## Verification
```bash
bun test tests/fpc.test.ts
# 4 pass, 0 fail
```

/claim #786

### 💰 Payout Details

### DIFF
--- a/src/fn/fpc.ts
+++ b/src/fn/fpc.ts
@@ -11,7 +11,7 @@ import { base_def } from "../helpers/zod/base_def"
 
 export const fpc_def = base_def.extend({
   fn: z.literal("fpc"),
-  num_pins: z.coerce.number().int().min(2),
+  num_pins: z.coerce.number().int().min(2).default(12),
   p: length.default("0.5mm").describe("contact pad pitch"),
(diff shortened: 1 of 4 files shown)


moxiaoshen567:
### 💰 Payout Details

PR #827, opened 2026-08-25
fix(pad): default dimensions to 1mm to prevent TypeError on bare name

### LINKED ISSUE #788: pad throws a raw TypeError from a bare name while smtpad defaults its dimensions

moxiaoshen567:
Fixes #788

## Summary
- Defaults `w` and `h` to `1mm` in `pad_def` (`src/fn/pad.ts`) and parses incoming params through `pad_def.parse` so calling `fp.string("pad")` renders a 1mm square pad (matching `smtpad` behavior) instead of throwing a raw `TypeError` from `@tscircuit/mm`.
- Added unit test in `tests/pad.test.ts` verifying bare string call produces a 1x1mm pad without throwing.

## Verification
```bash
bun test tests/pad.test.ts
# 3 pass, 0 fail
```

/claim #788

### 💰 Payout Details

### DIFF
--- a/src/fn/fpc.ts
+++ b/src/fn/fpc.ts
@@ -11,7 +11,7 @@ import { base_def } from "../helpers/zod/base_def"
 
 export const fpc_def = base_def.extend({
   fn: z.literal("fpc"),
-  num_pins: z.coerce.number().int().min(2),
+  num_pins: z.coerce.number().int().min(2).default(12),
   p: length.default("0.5mm").describe("contact pad pitch"),
   pw: length.default("0.3mm").describe("contact pad width"),
   pl: length.default("1.25mm").describe("contact pad length"),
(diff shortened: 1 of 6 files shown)


moxiaoshen567:
### 💰 Payout Details

PR #828, opened 2026-08-25
fix(to220): use standard 2.54mm lead pitch and respect p parameter

### LINKED ISSUE #790: to220 uses a body-width-derived pin pitch (2.6mm) instead of the standard 2.54mm

moxiaoshen567:
Fixes #790

## Summary
- Sets standard JEDEC 2.54mm (0.1 inch) default pitch in `to220_def` (`p: length.optional().default("2.54mm")`) and respects the `p` parameter instead of deriving lead pitch dynamically from body width (`maxHoleWidth / (numPins - 1)` which produced 2.6mm).
- Updated unit tests and snapshots in `tests/to220.test.ts` to verify standard 2.54mm hole coordinates (`x = -2.54, 0, 2.54`) and custom pitch support (`to220_3_p5mm`).

## Verification
```bash
bun test tests/to220.test.ts
# 8 pass, 0 fail
bun test tests/kicad-parity/to220f_3_kicad_parity.test.ts
# 1 pass, 0 fail
```

/claim #790

### 💰 Payout Details


moxiaoshen567:
### 💰 Payout Details

PR #830, opened 2026-08-25
fix(lcc): default pitch to standard 1.27mm to prevent pad overlap

### LINKED ISSUE #794: lcc footprints default to a 0.5mm pitch and emit overlapping pads

moxiaoshen567:
Fixes #794

## Summary
- Defaults `p` to `1.27mm` in `src/fn/lcc.ts` (matching JEDEC MO-047 / MS-034 standard for LCC/PLCC packages) so omitting an explicit pitch doesn't inherit generic 0.5mm quad pitch and cause 0.6mm-wide adjacent pads to overlap.
- Added unit test in `tests/lcc.test.ts` verifying bare `lcc44` defaults to 1.27mm pitch with positive clearance between neighboring pads.

## Verification
```bash
bun test tests/lcc.test.ts
# 2 pass, 0 fail
bun test tests/kicad-parity/lcc68_kicad_parity.test.ts
# 1 pass, 0 fail
```

/claim #794

### 💰 Payout Details


moxiaoshen567:
### 💰 Payout Details

PR #859, opened 2026-08-30
Set default footprint reference text size to 0.6mm

mohan-bee:
### Motivation
- Keep generated footprint reference text readable without overriding intentional custom sizes.

### Before
- Footprinter emitted 0.2mm for default reference text, which Core rendered at 0.4mm.

### After
- Footprinter emits 0.4mm by default, which Core renders at 0.6mm.
- Explicit sizes such as 0.3mm remain unchanged.
- 114 SVG snapshots, including 0402 and 0603, reflect only the intended font-size changes.


imrishabh18 (commented):
It is readable with `0.4mm` also, not sure why to increase the size? Where does this `0.6mm` number come from, is there a reference?

seveibar:
@imrishabh18 discussed on discord a bit, but kicad uses 0.8mm because jlcpcb and a lot of fabs don't have high quality silkscreen. 0.4mm is barely readable on the actual printed board, but is still technically readable. 0.6mm is a compromise to try to increase readability while still improving on the default density that kicad has

seveibar (commented):
i'm not seeing 0402 or 0603 snapshots update, i'm not sure this is comprehensive?

tests/kicad-parity/__snapshots__/0402_metric_parity.snap.svg
mohan-bee, line 1: this overlaps not sure this is ok ?

seveibar (approved):
i think you should bring the reference text a bit closer to the silkscreen stroke in some cases as well

@mohan-bee it might be a good idea, before merging to core, to have a followup that converts these to use realistic visual snapshots- e.g. using circuit-to-canvas and pngs. We should see how it looks in canvas/gerber to judge the sizing

PR #864, opened 2026-09-04
fix: honor SOT-23 builder pin counts

ardaerturk:
The fluent SOT-23 builder silently ignored its pin-count argument: `fp().sot23(5)`, `(6)`, and `(8)` all generated three pads, while the corresponding string definitions produced the correct footprints. Read the builder count when no count is supplied by the string, and expose the optional argument in the TypeScript signature. Unsupported counts now reach the existing validation instead of silently selecting three pins.

Validation: four regression cases fail before the fix; all 570 tests pass after it. The ESM/declaration build and changed-file Biome check pass. Added and visually reviewed SVG snapshots for the supported pin counts, as required by AGENTS.md.

AI-assisted with Codex; verified with runtime tests, geometry comparisons against the existing string API, and visual snapshots.


ardaerturk:
Following up on this September 4 fix for SOT-23 builder pin counts. The four recorded CI checks pass, and the PR includes the regression tests and SVG snapshots described above. Could a maintainer review it or let me know if a revision is needed? Please also confirm whether this contribution would be eligible for contributor rewards if merged; I have not assumed a fixed bounty.

PR #876, opened 2026-09-08
fix: preserve pin labels when hiding reference designators

nardeenal:
`norefdes` currently blanks every silkscreen text element, so using it on a `stampboard` or `stampreceiver` with pin labels also removes all pin identifiers.

Restrict the existing filter to the `{REF}` placeholder. Pin labels remain visible while the reference designator is hidden. Add regression coverage for both generated footprints with SVG snapshots.

Validation on macOS arm64 with Bun 1.4.2:
- Both new regression tests fail before the fix and pass after it.
- Focused tests: 10 passed.
- `bun test --timeout 30000`: 564 passed, two network-dependent tests timed out; both passed on a focused retry.
- `bun run build`, formatting of changed TypeScript files, and `git diff --check` pass.
- Both generated SVG snapshots were rendered and visually inspected.

Prepared and tested with OpenAI Codex assistance.


nardeenal:
I’m also interested in joining the contributor sponsorship program. Could a maintainer confirm whether new contributors can enroll now and whether PayPal is supported, or whether GitHub Sponsors must be enabled before contributions are merged? This contribution was prepared and tested with disclosed AI assistance. I understand this PR has no agreed fixed bounty and that merging it alone does not guarantee payment.

PR #889, opened 2026-09-18
fix: use the standard 2.54mm JEDEC pitch for to220 and honor the p parameter

Furox-Art:
/claim #790

## Problem

`to220` derived its lead pitch from the plastic body width:

```ts
const computedPitch = Math.max(2.5, (w * 0.4) / (numPins - 1))
```

- At the default `w=13mm`, 3 pins → **2.6mm**, 0.06mm off per step from the fixed JEDEC TO-220 pitch of 0.1in (2.54mm)
- The pitch shifted whenever body width changed, but a TO-220 lead pitch is fixed regardless of body size
- The declared `p` parameter (default `"5.0mm"`) was parsed and **never read** — `to220_p2.54mm` had no effect

## Fix

Mirror the sibling `to220f`, which already hardcodes `2.54mm` with the comment "to match KiCad":

- `p` defaults to `2.54mm` and is the single source of the pitch (`const pitch = mm(parameters.p)`)
- Body width no longer influences hole positions
- `parameters.p` reports the effective pitch

## Tests (6 new, snapshot-free, deterministic)

| Check | main | this PR |
|---|---|---|
| `to220_3` default pitch | 2.6mm | **2.54mm** |
| `p: "5mm"` override | ignored | honored → gaps [5, 5] |
| `w: "25.4mm"` body | pitch scales | still **2.54mm** |
| `to220_4` centered | ✓ | ✓ (unchanged) |
| `to220_5` pitch | derived | 2.54mm |
| reported `parameters.p` | computed | effective 2.54 |

## Snapshots

The 4 affected `to220_*.snap.svg` snapshots were regenerated **on ubuntu-latest CI** (via a temporary `workflow_dispatch` helper, removed before this PR) — no local-font drift can enter the pixel comparison. Every other snapshot in the repo is untouched (verified: diff touches exactly the 4 to220 SVGs).

Note: #828 also targets this issue but bundles fixes for three unrelated issues (#786/#788 fpc/pad defaults, solderjumper) into one PR; this PR is scoped to #790 only and includes the pitch-regression tests above.


Furox-Art:
Fresh local run of the regression suite (2026-09-20, bun 1.4.2, this branch):

```
bun test tests/to220-pitch.test.ts

 6 pass
 0 fail
 12 expect() calls
Ran 6 tests across 1 file. [1.26s]
```

Covers the 2.54mm TO-220 pitch geometry: pin pitch, pad size, body offsets and the regression snapshot for `to220("TO220_2")`.

